PR #15, opened 2020-01-08
feat: fix paging and missing parents

Caele:


src/data.js
```diff
@@ -8,7 +8,7 @@ export default {
       },
       measures: {
         min: 0,
-        max: 0,
+        max: 1,
```
haxxmaxx: what do we need a measure for?

Caele: it adds a name?

haxxmaxx: Ah ok. We use the third dimension for that currently, and the fourth for "details". Change the max for dimensions then?

Caele: currently it doesn't really matter, as we probably should use attribute expressions. Ok if I keep this here so that I can add measures without sense blocking me?

haxxmaxx: Sure, we will probably change this back and forth anyways

PR #16, opened 2020-01-09
feat: Use rectangle and text + other IE fixes

haxxmaxx:
Using the rectangle and text elements instead of foreignObject. Text manipulation needs to be done very manually, but totally doable. Managed to get ellipsis, shrinking text and wrapping to work in order to fit text inside the rectangle. I'm guessing UX wants wrapping but we know all three are possible.

Other text styling should be doable in a similar fasion as we styled the action button.


Caele:
Just add the work for the bbox as well, you can see my take on it here: https://github.com/qlik-oss/sn-org-chart/compare/tsm-chore/sense-create

PR #21, opened 2020-01-16
COLLECTED THINGS

Caele:
This is a draft for various things, but mainly the attribute expressions for coloring and labels. I will split these up into multiples before going ahead, but heres a sneak peak.

### DIFF
--- a/src/tree/render.js
+++ b/src/tree/render.js
@@ -23,7 +23,12 @@ const filterTree = (id, tree) => {
       currentNode.data.id === node.data.id ||
       (currentNode.parent && node.data.id === currentNode.parent.data.id) ||
       (currentNode.parent && node.parent && node.parent.data.id === currentNode.parent.data.id) ||
-      (node.parent && node.parent.data.id === currentNode.data.id)
+      (node.parent && node.parent.data.id === currentNode.data.id) ||
+      (currentNode.parent &&
+        node.parent &&
+        node.parent.parent &&
+        node.parent.parent.data.id === currentNode.parent.data.id &&
+        !node.children)
     );
   });
 };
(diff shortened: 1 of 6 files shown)


src/tree/render.js
```diff
@@ -23,7 +23,12 @@ const filterTree = (id, tree) => {
       currentNode.data.id === node.data.id ||
       (currentNode.parent && node.data.id === currentNode.parent.data.id) ||
       (currentNode.parent && node.parent && node.parent.data.id === currentNode.parent.data.id) ||
-      (node.parent && node.parent.data.id === currentNode.data.id)
+      (node.parent && node.parent.data.id === currentNode.data.id) ||
+      (currentNode.parent &&
```
haxxmaxx: Show the current node's sibling's children? Nieces I guess 😅

src/utils/tree-transform.js
```diff
@@ -72,6 +74,35 @@ const getDataMatrix = async (layout, model) => {
...
+}
+
+function getAttribute(attrs, index) {
+  if (index !== false) {
+    return attrs.qValues[index].qText;
+  }
+  return false;
+}
+
+function transformColor(color) {
```
haxxmaxx: We have a bunch of stuff in the button repo handling color, if it's needed :)

src/tree/render.js
Caele: yea, this part is not working properly anyway. Basically I wanted to show all subtrees without children as vertical stacks, but it breaks down for un-even trees
image

src/utils/tree-transform.js
Caele: yea, basically I just solved the ARGB case that engine spits out, but there are likely more things. And if we already solved them, we could pick them from there. 
This is probably something that should end up in nebula later.

haxxmaxx: ah true that would be really nice

PR #24, opened 2020-01-20
feat: change default paths to rounded elbow

eliseeborn:
rounded elbow paths, should work regardless of the two points relation to each other. rdef can be changed to get larger/smaller curve on the corner.


haxxmaxx:
Looks so much better. I'd say let's merge this right away and I can try if it works with all orientation.

PR #25, opened 2020-01-21
feat: all four orientations

haxxmaxx:


src/tree/render.js
```diff
@@ -102,65 +120,188 @@ const reRenderTree = ({ svg, activeNode, allNodes, o, width, height }) => {
...
-              (d.data.childNumber - d.parent.children.length / 2) * (nodeSize.width + spaceBetweenNodes)
-            : center;
-        return d.xActual;
-      },
-      y: function(d) {
-        return d.y;
-      },
-    },
-  };
+  let orientations;
```
haxxmaxx: Not sure this is the best way to do it. I guess you can have multiple options inside the the orientations object, but that requires rewriting a lot of other stuff. @Caele thoughts?

eliseeborn:
zoomningen in på trädet är off i ltr & rtl. 
Screenshot 2020-01-21 at 08 23 48
annars ser det bra ut tycker jag

haxxmaxx:
> zoomningen in på trädet är off i ltr & rtl.
> Screenshot 2020-01-21 at 08 23 48
> annars ser det bra ut tycker jag

There is a todo comment for that. But I just got an idea on how to fix it

PR #26, opened 2020-01-21
feat: view more

eliseeborn:
just removing children/siblings beyond a certain point, need to add something to actually show the hidden ones as well

### DIFF
--- a/src/tree/render.js
+++ b/src/tree/render.js
@@ -7,8 +7,10 @@ import '../treeCss.css';
 const nodeSize = { width: 300, height: 100 };
 const siblingSpacing = 30;
 const transitionTime = 500;
-const orientation = 'rtl';
+const orientation = 'ttb';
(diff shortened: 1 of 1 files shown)


src/tree/render.js
```diff
@@ -231,7 +233,7 @@ const renderTree = async ({ element, layout, app, model }) => {
                 ? d.parent.xActual +
                   nodeSize.width / 2 +
                   siblingSpacing / 2 +
-                  (d.data.childNumber - d.parent.children.length / 2) * (nodeSize.width + siblingSpacing)
+                  (d.data.childNumber - (d.parent.children.length < childrenToShow ? d.parent.children.length : childrenToShow) / 2) * (nodeSize.width + siblingSpacing)
```
haxxmaxx: I see now that we probably should break this function out and and send xActual or yActual in as arguments. So we only have to make changes to one function in the future.

PR #27, opened 2020-01-22
feat: refactor code and replace foreignobject

Caele:


src/tree/render.js
```diff
@@ -325,7 +176,7 @@ const renderTree = async ({ element, layout, app, model }) => {
     // Using the treemap created
     const allNodes = treemap(nodes);
     const activeNode = allNodes.data.id;
-    reRenderTree({ svg, data, allNodes, activeNode, o, width, height, treemap });
+    reRenderTree({ svg, divBox, data, allNodes, activeNode, o, width, height, treemap });
```
haxxmaxx: Do we need to send all these? `allNodes` and `activeNode` might as well be created inside `reRenderTree` so we can pass fewer arguments?

src/treeCss.css
```diff
@@ -21,6 +21,7 @@ body {
 .nodeRect {
   border: 2px solid black;
   background: transparent;
+  position: absolute;
 }
 
 .nodeReact .text {
```
haxxmaxx: I guess this is not used anymore?

src/tree/render.js
```diff
@@ -182,124 +142,7 @@ const renderTree = async ({ element, layout, app, model }) => {
...
-          x: function(d) {
-            return d.y;
-          },
-        },
-      };
-      break;
-    default:
-      break;
-  }
+  const orientations = position(orientation, nodeSize);
```
haxxmaxx: The naming here is a bit weird. I don't know if we still want call it `orientations`, since it will only contain one orientation the way I refactored it. At least, I think this line should be something like `someName = getSomeName(...)`.

src/tree/render.js
Caele: they are passed along when you re-render the tree, so they still need to be arguments. But I think that part can be optimized somehow, especially when I looked at selections. But for now, they can stay.

src/treeCss.css
Caele: probably not, but text will be added in, so can keep it

src/tree/render.js
Caele: yea maybe, I renamed these quite a few times before just picking one. I liked the simplicity of position, path, box... orientation didn't make much sense as it is position data you get back.
For "position" you can read it as a verb, "to position something", which makes more sense. but then, "to box something".... anyway, can always change it later, important part was to separate them.

haxxmaxx: Ok, yeah better naming tend to pop up along the way :)

src/treeCss.css
haxxmaxx: Ahh, maybe it is just a typo? I thought it was something related to react

Caele: ah you mean that, probably just a left over or typo

PR #28, opened 2020-01-22
test: add basic

Caele:


.eslintrc.json
```diff
@@ -8,17 +8,19 @@
   },
   "extends": ["airbnb-base"],
   "rules": {
-    "arrow-parens": 2,
-    "operator-linebreak": 0,
+    "arrow-parens": 0,
+    "comma-dangle": 0,
+    "implicit-arrow-linebreak": 0,
     "max-len": 0,
```
haxxmaxx: I'd say we want comma-dangle and some max-len. The rest looks good

Caele: yea, I agree, but there is something with them that breaks,

PR #29, opened 2020-01-23
feat: render leaf only branches orthogonal

haxxmaxx:
A lot of math here... Tried to parameterize as much as possible.
The paths will always be to the left for vertical trees and to on top for horizontal trees.
The margin between them is enough to show we could show multiple branches, whithout rerendering the parents 👍 
Screenshot 2020-01-23 at 15 58 40
Screenshot 2020-01-23 at 16 14 00

A couple of concepts that would be nice to define for the chart, if we need to use them again:
- **depth**, the direction of the depth of the tree (e.g. top-to-bottom, this is the y axis)
- **width**, the opposite direction of depth. (so the x-axis for a top-to-bottom tree)


src/utils/util.js
```diff
@@ -0,0 +1,8 @@
+export default function isOnlyLeafs(children) {
```
haxxmaxx: Not the best name, but I think we should have some general util file

src/tree/box.js
```diff
@@ -1,8 +1,8 @@
 import card from '../card/card';
 
-export default function box(divBox, o, nodeSize, appendNodes, reRender) {
+export default function box(divBox, o, appendNodes, reRender) {
```
niekvanstaveren: We should rename o here as well. I copied this from an example but it is not really clear what it represents right now. Will think of something

src/utils/util.js
niekvanstaveren: hasGrandChildren ?

niekvanstaveren: or in this way hasNoGrandChildren I guess

eliseeborn: onlyLeafChildren maybe? or hasOnlyLeafChildren?

haxxmaxx: Ah... I was referring to the file name 😅The function name is pretty accurate imo. We are passing the children array and the function returns true if they all are leafs.

src/tree/box.js
haxxmaxx: Agree. It probably is short for orientation. It contains the object we return from the `position` file. Had the same discussion with Tobias. But will figure out something eventually :)

src/utils/util.js
niekvanstaveren: Ah okay. Maybe tree-utils? In that case we could add this function also to the tree-transform util file and rename that one

haxxmaxx: Sounds good, I'll do that

src/tree/path.js
```diff
@@ -1,51 +1,75 @@
...
+          return `
+            M ${start.x} ${start.y}
+            L ${start.x} ${end.y - halfNode.y - r.y}
+            C ${start.x} ${end.y - halfNode.y} ${start.x} ${end.y - halfNode.y} ${start.x + r.x} ${end.y - halfNode.y}
+            L ${end.x + halfDepth - r.x} ${end.y - halfNode.y}
+            C ${end.x + halfDepth} ${end.y - halfNode.y} ${end.x + halfDepth} ${end.y - halfNode.y} ${end.x + halfDepth} ${end.y - halfNode.y - r.y}
+            L ${end.x + halfDepth} ${end.y + r.y}
+            C ${end.x + halfDepth} ${end.y} ${end.x + halfDepth} ${end.y} ${end.x + halfDepth + r.x} ${end.y}
+            L ${end.x} ${end.y}
+            `;
```
Caele: Should be able to name these two cases and move them to their own function/file.

Caele:
Think it looks ok, without having delved into the maths. Add some unit tests that shows what the calculations are and then a story/task to add integration tests as well. Don't think you can add the tests now as you'd need to a property for the currentNode to make it fully.

src/tree/path.js
haxxmaxx: I actually just had an idea of how to generalize the whole thing. I'll try that otherwise I'll brake out those cases somehow

haxxmaxx:
> Looks good, I think 👍
> This is a pretty funny corner case I found, but I don't think changing it to a straight line is necessary

Hmm yeah maybe can add a case for that later, if Christofer wants it or something. I'll merge for now

PR #31, opened 2020-01-24
feat: styling options

niekvanstaveren:
Only draft PR right now. Will need some input from UX to get things right. However would be great if someone could have a quick look. 

The changes in card.js I'm definitely not sure about yet, however those will definitely need more changes when we have more input


src/utils/styling.js
```diff
@@ -0,0 +1,32 @@
...
+
+    const detailsColor = layout.style.details.useColorExpression
+      ? colorUtils.resolveExpression(layout.style.details.colorExpression)
+      : colorUtils.resolveColor(layout.style.details.color, palette);
+
+    const cardStyling = {
+      backgroundColor,
+      label: `white-space:nowrap;overflow:hidden;text-overflow:ellipsis;font-size:24px;color:${labelColor}`,
+      subLabel: `white-space:nowrap;overflow:hidden;text-overflow:ellipsis;font-size:20px;color:${subLabelColor}`,
+      details: `white-space:nowrap;overflow:hidden;text-overflow:ellipsis;font-size:20px;color:${detailsColor}`,
```
Caele: Should move constants to css file.

src/card/card.js
```diff
@@ -1 +1,13 @@
-export default (data) => `<div class="orgCard">${data.name}</div>`;
+export default (data, cardStyling) => {
+  if (!data.attributes) {
+    return null;
+  }
+  let html = `<div style=${cardStyling.label}>${data.attributes.label}</div>`;
```
Caele: if label is missing, you should fallback to ID

src/utils/styling.js
```diff
@@ -0,0 +1,32 @@
+import colorUtils from './color-utils';
+
+const stylingUtils = {
+  cardStyling: ({ Theme, layout }) => {
+    const palette = colorUtils.getPalette(Theme);
+    const backgroundColor = layout.style.background.useColorExpression
```
haxxmaxx: Could make this into a function, passing 'background', 'label' etc as a key.

PR #37, opened 2020-01-28
fix: disable transitions in ie11

Caele:
Turns out the transitions in ie11 is quite hard to fix, and it would make it worse in other browsers. My current thought is to just check for ie11 and disable those transitions. Basically not spend more time on it.


niekvanstaveren:
> Turns out the transitions in ie11 is quite hard to fix, and it would make it worse in other browsers. My current thought is to just check for ie11 and disable those transitions. Basically not spend more time on it.

Seems like the smart way to go. Is this ready for review?

PR #40, opened 2020-01-31
chore: update nebula and use hooks

Caele:
Update to latest Nebula version and refactor a bit to use hooks.


src/index.js
```diff
@@ -10,21 +12,40 @@ export default function supernova(env) {
...
-      willUnmount() {},
-      destroy() {},
+    component: () => {
+      //  const [focusNodes, setFocusNodes] = useState([]);
+      const [dataTree, setDataTree] = useState({});
+      //   const [storage, setStorage] = useState({});
+      const layout = useStaleLayout();
+      const model = useModel();
+      const element = useElement();
+      const rect = useRect();
```
haxxmaxx: bounding rect?

Caele: basically the size of the chart

package.json
```diff
@@ -46,7 +46,7 @@
     ]
   },
   "peerDependencies": {
-    "@nebula.js/supernova": "0.1.0-alpha.25"
+    "@nebula.js/supernova": "0.1.0-alpha.28"
```
miralemd:
```suggestion
    "@nebula.js/supernova": ">=0.1.0-alpha.28"
```

src/index.js
```diff
@@ -10,21 +12,40 @@ export default function supernova(env) {
...
+
+      const callRender = () => {
+        if (element && dataTree && layout) {
+          return renderTree({
```
miralemd: A function returned from `useEffect` is a 'cleanup' function that will be executed when the specified deps change, and when the component is unmounted/destroyed - you probably need to split this method for usePromise and useEffect

src/index.js
```diff
@@ -10,21 +12,40 @@ export default function supernova(env) {
...
+            element,
+            dataTree,
+            layout,
+            Theme,
```
miralemd: `env.Theme` is considered deprecated - `useTheme` instead

src/index.js
Caele: think I want to use usePromise for all calls to render(), as they are likely to be async. Not needed now thought, and got into some infinite loops with it calling the rect one all the time.

PR #41, opened 2020-01-31
fix: Rerender all nodes

haxxmaxx:
Current flow:

**treeTransform** runs when `layout` or `model` is updated
- creates dataTree
- gets styling

**preRenderTree** runs when `element` or `dataTree` is updated + resizing
- create `svg`, `divBox`, `positioning` (prev `o`), `width` and `height`
- returns them to index and saved as `objectData` state

**paintTree** runs when activeNode is updated,
- remove old stuff from `divBox` and `svg`
- `filterTree`
- create boxes, paths
- `transform`


src/utils/styling.js
```diff
@@ -18,8 +18,8 @@ export function getColor(reference, palette, defaultColor) {
 const stylingUtils = {
   cardStyling: ({ Theme, layout }) => {
     const palette = colorUtils.getPalette(Theme);
-    const backgroundColor = getColor(layout.style.backgroundColor, palette, '#e6e6e6');
-    const fontColor = getColor(layout.style.fontColor, palette, 'default');
+    const backgroundColor = getColor(layout.style.background, palette, '#e6e6e6');
+    const fontColor = getColor(layout.style.label, palette, 'default');
```
niekvanstaveren: Don't think you can change these only here. Does the styling still works?

src/tree/render.js
```diff
@@ -31,62 +28,33 @@ const filterTree = (id, nodeTree) => {
...
-  }
-  previousNodes = nodes;
+  divBox.selectAll('.nodeRect').remove();
+  svg.selectAll('g').remove();
 
   // Create the nodes.
   const node = svg
     .selectAll('.node')
-    .data(appendNodes)
+    .data(nodes)
```
niekvanstaveren: Aren't we attaching also existing nodes again in this case?

src/utils/styling.js
haxxmaxx: It was complaining that backgroundColor did not exist on the layout. So looked and they were called background and label. Is that old stuff?

niekvanstaveren: Possibly because you were using an older chart. I moved it to backgroundColor and fontColor since it is only containing color info

haxxmaxx: ok I'll reset this

src/tree/render.js
niekvanstaveren: Wait we are deleting everything and building up everything again as well right?

haxxmaxx: Yes, we are. I'm unsure if we need to pass the svg and divBox around. Since everything in them is removed, maybe create them inside rerender? @Caele thoughts?

Caele: they only need to be created if the element changes really, so in theory they should only be created once, then we store a reference to them. 
Using the hooks we should be able to separate things into:
- loadData
- init elements
- construct tree
- render tree

This is fine for now though, it's the most straightforward way, not complicating things.

src/tree/render.js
```diff
@@ -27,84 +26,61 @@ const filterTree = (id, nodeTree) => {
...
-      activeNode: id,
-      width,
-      height,
-      cardStyling,
-      callback,
-      storage,
-    });
+  // Create cards
+  box(divBox, positioning, nodes, styling, id => {
+    setActiveCallback(id);
```
niekvanstaveren: I guess this behaviour will change a bit with the changes from Elise, but this looks already nicer I would say

src/index.js
```diff
@@ -55,16 +67,19 @@ export default function supernova(env) {
...
         return Promise.resolve();
       }, [layout, model]);
 
-      // Should avoid dependency on layout -> store that on dataTree instead?
-      // Should move element dependency to preRender function
-      useEffect(callRender, [layout, element, dataTree]);
-      // Should be bundled with dataTree dependency once layout and element has been moved?
-      useEffect(callRender, [rect]);
+      // This one can split up. Only need to get new height/width when rect is changed
+      useEffect(callPreRender, [element, dataTree, rect.width, rect.height]);
```
niekvanstaveren: is putting just rect here not enough?

src/tree/render.js
```diff
@@ -143,45 +89,17 @@ const renderTree = ({ element, dataTree, layout, Theme }) => {
 
   const divBox = select(element)
     .selectAll('div')
-    .data(entries(orientations))
+    .data([{}])
```
niekvanstaveren: Is attaching data necessary if it is empty?

haxxmaxx: Tried without it, didn't work.

src/index.js
haxxmaxx: hmm it got caught in an endless loop before, but works using only rect now

src/tree/render.js
haxxmaxx: Yes. We should send stuff back to index here and then run an update from there.

PR #42, opened 2020-02-04
feat: selections

niekvanstaveren:


src/index.js
```diff
@@ -29,6 +31,40 @@ export default function supernova(env) {
...
+        () => ({
+          action() {
+            setLinked(!linked);
+          },
+          icon: {
+            d: 'svg path',
+          },
+        }),
+        [linked]
+      );
```
niekvanstaveren: This should create a button on the selection toolbar. However not supported by Nebula yet. The svg path needs to be changed of course to the icon that we are going to use

src/index.js
```diff
@@ -29,6 +31,40 @@ export default function supernova(env) {
...
+          }
+        };
+        document.addEventListener('keydown', addKeyPress);
+        document.addEventListener('keyup', removeKeyPress);
+
+        return () => {
+          document.removeEventListener('keydown', addKeyPress);
+          document.removeEventListener('keyup', removeKeyPress);
+        };
+      }, []);
```
niekvanstaveren: Since there is no button for chained selection yet I put this in place so we can test this. We might keep this in, but need to discuss with UX I guess.

haxxmaxx:
Oh boy we might have some interesting conflicts... I'll get what I have working up and open a PR so we can sync about this.

src/treeCss.css
```diff
@@ -26,6 +26,17 @@
   padding-left: 5px;
   padding-right: 5px;
   border-radius: 0px 0px 3px 3px;
+  -webkit-touch-callout: none;
+  -webkit-user-select: none;
+  -khtml-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  user-select: none;
```
niekvanstaveren: So I got annoyed that text was selected when I tried to make selections. Not sure if we should keep this. Again need to ask UX here I guess

niekvanstaveren:
> Oh boy we might have some interesting conflicts... I'll get what I have working up and open a PR so we can sync about this.

Oh no. I was expecting conflicts between what you are working on and what Elise is doing not for this one! Yes let's sync

haxxmaxx:
> Oh no. I was expecting conflicts between what you are working on and what Elise is doing not for this one! Yes let's sync

Actually should not be that bad. It justs the passing of linked and selectionAPI that needs to be changed I think.

src/treeCss.css
Caele: yea we generally don't allow text selection, so this should be fine

src/utils/selections.js
```diff
@@ -0,0 +1,57 @@
...
+
+      if (selections.selectionState.length === 0) {
+        selections.clear();
+      } else {
+        selections.select({
+          method: 'selectHyperCubeValues',
+          params: ['/qHyperCubeDef', 0, selections.selectionState, false],
+        });
+      }
+    }
```
Caele: After all this you should trigger an internal re-render right?

niekvanstaveren: I guess when you clear the selections you should do a re-render to clear up the styling of selected nodes. However when you make a selection the layout should change and this should be picked up in the useEffect methods. Would be the same when you make selections outside the org chart.

Caele: inside the client the layout actually doesn't change, there is no new paint call for charts in the selection state.
I think we can create a function that only updates the divs, and in the long run only updated certain divs, filtered on their changed selection state.

src/utils/selections.js
```diff
@@ -0,0 +1,60 @@
+import { getAllTreeElemNo } from './tree-utils';
+
+export const selectionState = {
+  linked: false,
+};
```
Caele: won't this become a shared state between org charts on the same sheet?

niekvanstaveren: Yes probably, however sending down linked through the layers won't work. Only when things re-render the new state would then be passed down.

src/index.js
```diff
@@ -32,6 +35,41 @@ export default function supernova(env) {
       const element = useElement();
       const rect = useRect()[0];
       const Theme = useTheme();
+      const [linked, setLinked] = useState(false);
```
haxxmaxx: My OCD wants this next to the other state declarations 🙃

src/tree/box.js
```diff
@@ -17,9 +18,9 @@ export default function box(divBox, o, appendNodes, cardStyling, reRender) {
     .attr('style', getStyle)
     .attr('id', d => d.data.id)
     .on('click', node => {
-      if (node.children) {
-        reRender(node.data.id);
+      if (node.data.id !== 'Root') {
```
haxxmaxx: The old logic for traversing the tree will also be called from here. But this is fine. We will distinguish between them by looking at the target of the click event

niekvanstaveren: Ah okay. You are not creating a new div?

src/utils/selections.js
niekvanstaveren: Put this on the selectionsAPI now as well

src/tree/box.js
haxxmaxx: I am creating a new div now :)

PR #43, opened 2020-02-04
fix: hide expressions for second dimension

Caele:


niekvanstaveren:
Cool! Didn't look into if this was possible, but that would look a lot better 👍

PR #44, opened 2020-02-05
feat: show third label or measure on card

eliseeborn:


src/extension/ext-raw.js
```diff
@@ -41,6 +41,14 @@ export default {
                     id: 'subLabelExpression',
                     tid: 'subLabelExpression',
                   },
+                  {
+                    component: 'expression',
+                    ref: 'qExpression',
+                    translation: '$Extra label expression',
```
niekvanstaveren: We shouldn't forget to get in good translations here. Luckily we have it on our mindmap :)

PR #46, opened 2020-02-06
feat: add tooltips for cards

eliseeborn:
some basic tooltips containing all the card data
I know we have tooltips in sense, these should probably look the same as well


src/card/card.js
```diff
@@ -34,5 +34,6 @@ export default (data, cardStyling) => {
   } else if (attributes.extraLabel) {
     html += `<div class="org-card-text">${attributes.extraLabel}</div>`;
   }
+  html += `<div data-html="true" class="tooltip">${attributes.label || data.id}</br>${attributes.subLabel || ''}</br>${attributes.extraLabel || ''}</br>${data.measure || ''}</div>`;
```
Caele: This will add one tooltip per card, would be better to just add a single one (appended to the divBox or something) that you then move around.

`
tooltip = divBox.append('div')
    .attr('class', 'tooltip')
    .attr('style', p => getMoreStyle(p, true))
    .on('mouseover', () => {
     tooltip.append("SOME NICE TEXT OR WHAT?")..classed('show-tooltip', true)
    }).on('mouseout'()=> {tooltip.classed(('show-tooltip', false)});
`

src/tree/box.js
```diff
@@ -8,6 +8,17 @@ export default function box(divBox, o, appendNodes, cardStyling, reRender) {
     return `width:${o.nodeSize.width}px;height:${o.nodeSize.height}px; top:${o.y(p)}px;left:${o.x(p)}px;`;
   }
 
+  function getTooltipStyle(p) {
+    if (p.data.id === 'Root') {
+      return `top:${o.y(p) + 55}px;left:${o.x(p) + 140}px;visibility: visible; opacity: 1;`;
+    }
+    return `width:200px;height:50px; top:${o.y(p) - 25}px;left:${o.x(p)}px;visibility: visible;opacity: 1;`;
```
eliseeborn: this style is not applied, so the tooltip does not move to the right position right now

src/tree/box.js
```diff
@@ -21,5 +32,16 @@ export default function box(divBox, o, appendNodes, cardStyling, reRender) {
         reRender(node.data.id);
       }
     })
-    .html(d => card(d.data, cardStyling));
+    .html(d => card(d.data, cardStyling))
+    .on('mouseover', d => {
+      tooltip
+        .html(`${d.data.attributes.label || ''}<br />${d.data.attributes.subLabel || ''}<br />${d.data.attributes.extraLabel || ''}<br />${d.data.measure || ''}<br />`)
```
eliseeborn: this results in too many <br> if any attributes are missing

src/tree/box.js
```diff
@@ -21,5 +32,16 @@ export default function box(divBox, o, appendNodes, cardStyling, reRender) {
         reRender(node.data.id);
       }
     })
-    .html(d => card(d.data, cardStyling));
+    .html(d => card(d.data, cardStyling))
+    .on('mouseover', d => {
+      tooltip
+        .html(`${d.data.attributes.label || ''}<br />${d.data.attributes.subLabel || ''}<br />${d.data.attributes.extraLabel || ''}<br />${d.data.measure || ''}`)
+        .attr('style', 'visibility: visible;opacity: 1;')
+        .attr('style', () => { getTooltipStyle(d); });
```
Caele: Missing return here to get the style to apply.

src/tree/box.js
```diff
@@ -21,5 +32,16 @@ export default function box(divBox, o, appendNodes, cardStyling, reRender) {
         reRender(node.data.id);
```
Caele: close tooltips on click as well, else when the tree re-renders you'll never get a mouseout

src/tree/render.js
```diff
@@ -94,12 +95,19 @@ export function preRenderTree(element, dataTree) {
     .append('div')
     .attr('class', 'org-node-holder');
 
+  const tooltip = select(element)
+    .selectAll('.tooltip')
```
Caele: change classname to something more specific: `sn-org-tooltip`

Caele: Also update from master to get the css changes, and create a new file for the tooltip css/less.

eliseeborn:
the only things now that maybe should be fixed:
- the tooltips are a constant size, so if there are very long attributes the text might overflow (but then the positioning will get tricky again)
- they don't close after a certain time has passed like in sense
- renaming the selections object, which I will do now

src/tree/box.js
```diff
@@ -74,6 +75,12 @@ export default function box(
     .attr('style', d => `top:${y(d) - rootDiameter - buttonMargin}px;left:${x(d) + (cardWidth - rootDiameter) / 2}px`)
     .attr('id', d => d.data.id);
 
+  function getTooltipStyle(d) {
+    const halfCardWidth = cardWidth / 2;
+    const halfTooltipWidth = 240 / 2;
```
niekvanstaveren: We should have tooltip size also in the constants file

src/tree/box.js
```diff
@@ -74,6 +75,12 @@ export default function box(
...
+    return `top:${(y(d)) * sel.zoom.zoom + sel.zoom.y - 85}px;left:${x(d) * sel.zoom.zoom + sel.zoom.x - (halfTooltipWidth - (halfCardWidth * sel.zoom.zoom))}px;visibility: visible;opacity: 0.9;`;
```
niekvanstaveren: zoom.zoom nice. Not sure if Max approves though :P

src/tree/box.js
```diff
@@ -88,7 +95,24 @@ export default function box(
...
+    .on('mouseleave', () => {
+      tooltip
+        .html('')
+        .attr('style', 'visibility: hidden;opacity: 0;');
+    })
+    .on('mousedown', () => {
+      tooltip
+        .html('')
+        .attr('style', 'visibility: hidden;opacity: 0;');
+    });
```
niekvanstaveren: This does not interfere with zoom/panning or making selections?

yarn.lock
```diff
@@ -1960,11 +1960,6 @@ JSONStream@^1.0.4:
     jsonparse "^1.2.0"
```
niekvanstaveren: I also had differences in my yarn.lock file. I guess we forgot to commit it somewhere

src/index.js
```diff
@@ -47,7 +48,7 @@ export default function supernova(env) {
       const [opts] = useState(options);
       const selectionsAPI = useSelections();
       const constraints = useConstraints();
-      const [selections] = useState({ api: selectionsAPI, setState: setSelectionState, linked: false });
+      const [selections] = useState({ api: selectionsAPI, setState: setSelectionState, linked: false, zoom: {} });
```
niekvanstaveren: Let's go for selectionsAndZoom if we cannot come up with anything better for now

src/tree/box.js
eliseeborn: This is mostly so that the tooltip will close when panning, though they can reopen, so it isn't very effective. The panning and selections do not seem to be affected.

yarn.lock
eliseeborn: should I remove it from the pr, or leave it?

niekvanstaveren: Keep it in, hopefully this will fix things for everyone :)

src/index.js
```diff
@@ -47,30 +48,34 @@ export default function supernova(env) {
       const [opts] = useState(options);
       const selectionsAPI = useSelections();
       const constraints = useConstraints();
-      const [selections] = useState({ api: selectionsAPI, setState: setSelectionState, linked: false });
+      const [selectionsAndZoom] = useState({ api: selectionsAPI, setState: setSelectionState, linked: false, transform: {} });
```
haxxmaxx: If we store both zoom and translation, it should be `selectionsAndTransform` I guess

src/tree/box.js
```diff
@@ -83,6 +84,12 @@ export default function box(
     .attr('style', d => `top:${y(d) - rootDiameter - buttonMargin}px;left:${x(d) + (cardWidth - rootDiameter) / 2}px`)
     .attr('id', d => d.data.id);
 
+  function getTooltipStyle(d) {
+    const halfCardWidth = cardWidth / 2;
+    const halfTooltipWidth = tooltipWidth / 2;
+    return `top:${(y(d)) * sel.transform.zoom + sel.transform.y - tooltipHeight - 25}px;left:${x(d) * sel.transform.zoom + sel.transform.x - (halfTooltipWidth - (halfCardWidth * sel.transform.zoom))}px;visibility: visible;opacity: 0.9;`;
```
haxxmaxx: Since we are not using `tooltipHeight` anywhere else atm, set it to 60 + 25 and remove -25 here.

src/tree/box.js
```diff
@@ -97,7 +104,24 @@ export default function box(
         selections.select(node, sel, selectionState);
       }
     })
-    .html(d => card(d.data, cardStyling, sel, selectionState));
+    .html(d => card(d.data, cardStyling, sel, selectionState))
+    .on('mouseenter', d => {
+      if (allowInteractions) {
```
haxxmaxx: Actually we should use `!constraints.passive` here. Tooltips should be shown play mode for snapshots. And passive is undefined in play mode, where active is true

src/styles/treeCss.less
```diff
@@ -0,0 +1,38 @@
+.sn-org-chart {
```
Caele: Rename this file: tooltip.less

PR #50, opened 2020-02-07
feat: zoom and pan mode

niekvanstaveren:
So this would work when you set up zooming and panning from the start. However it does not consider state at all. So not sure what would happen if you have an active node different from the top node and resize. Will have to merge in the changes from Max to test that. However here a quick peak

Also would need to add something in PP to have the option to change between the modes

TODO:
- [ ] Keep correct state when switching between selections (and size)
- [ ] Filter on zoom actions 
- [ ] Limit on number of children (100)
- [ ] Tooltip should be still attached to card
- [ ] Keep showing parents when expanding downwards
- [ ] Store zoom and pan state for Viewstate


src/tree/render.js
```diff
@@ -60,10 +62,10 @@ export const getSize = ({ error, warn }, element) => {
   return size;
 };
 
-export function preRenderTree(element, dataTree) {
+export function preRenderTree(element, dataTree, layout) {
```
haxxmaxx: We tried to make preRender independent of layout, not sure we want it here. Is there a way around it?

niekvanstaveren: Yes should be possible. Thought the mode of zooming would be nice to have on the objectData. I will have a look

src/extension/ext-raw.js
```diff
@@ -101,6 +107,19 @@ export default {
...
+                      navigtaionMode: {
+                        ref: 'navigationMode',
+                        type: 'string',
+                        translation: '$Navigationmode',
+                        component: 'dropdown',
+                        default: 'scroll',
+                        options: navigationOptions,
+                      },
+                    },
+                  },
```
Caele: Add this stuff in, but for now set default to free and hide the option. That way we can publish the chart without exposing the other options for now.

src/tree/render.js
Caele: Doesn't really matter as preRender is already reliant on dataTree. Just avoid adding layout as a dependency to the hook.
you could instead set it as a property on the root node, that way you'll have access to it, I did that with the warnings.

src/extension/ext-raw.js
niekvanstaveren: We want to allow for automatic as well right? Should be able to set that in property panel I guess

src/tree/render.js
niekvanstaveren: Moved it to the root node

PR #52, opened 2020-02-11
fix: Path adjustment and expand multiple children whenever possible

haxxmaxx:
Branched of my old branch, hence all the commits...

- Set all sizes as constants
- Fix paths for to buttons. **This breaks support for others directions than top-to-bottom.** Should not be that hard to fix later.


src/tree/size-constants.js
```diff
@@ -0,0 +1,17 @@
...
+const buttonHeight = 24;
+
+const constants = {
+  cardWidth: 152,
+  cardHeight: 64,
+  widthMargin: 32,
+  heightMargin: buttonHeight + buttonMargin * 3,
+  leafMargin: 16,
+  buttonMargin,
+  buttonWidth: 48, // TODO: might need to be dynamic
```
Caele: you could probably just skip this, and instead set a max-width in css, just to avoid strange large cases.

haxxmaxx: Exactly, to center them becomes a problem if the css sets the width dynamically. I've tried and three digits fit, so we'll never need more than that if we have a sanity limit.

PR #57, opened 2020-02-12
Tsm/css select

Caele:
Started with some css, ended up with refactoring selections a bit.
- Added less support to clean up classes
- divided the css into multiple files
- changed selectors to be more specific
- Removed some surplus elements, like extra divs holding paths and cards
- Removed all instances of setting objects to the selections API
- Correctly remove listeners to selections API using hook return function
- updated nebbie


src/styles/paths.less
```diff
@@ -0,0 +1,17 @@
+.sn-org-chart {
+  .sn-org-svg {
+    position: absolute;
+
+    .sn-org-link {
```
haxxmaxx: I think we should call them either links or paths, not both

Caele: yea, paths it is

PR #61, opened 2020-02-13
fix: resize svg when resizing object

niekvanstaveren:


niekvanstaveren:
Any idea on how to do this better @Caele. We need to resize the svg element otherwise the paths are not shown outside. 
This happens when you make the chart bigger but the inner svg keeps the previous size

Caele:
> Any idea on how to do this better @Caele. We need to resize the svg element otherwise the paths are not shown outside.
> This happens when you make the chart bigger but the inner svg keeps the previous size

Think you can change the svg to be 100%/100% sized instead.

niekvanstaveren:
> > Any idea on how to do this better @Caele. We need to resize the svg element otherwise the paths are not shown outside.
> > This happens when you make the chart bigger but the inner svg keeps the previous size
> 
> Think you can change the svg to be 100%/100% sized instead.

Ah yes that does work. I thought I had tried that before

PR #65, opened 2020-02-13
feat: show parent tree for zoom mode

niekvanstaveren:


src/tree/render.js
```diff
@@ -19,6 +19,14 @@ const filterTree = ({ topId, isExpanded, expandedChildren }, nodeTree) => {
...
     });
   }
+
+  if (nodeTree.data.navigationMode === 'free' && topNode.parent) {
+    const ancestors = topNode.parent.ancestors();
+    ancestors.forEach(ancestor => {
+      subTree.push(ancestor);
+      subTree.push(...ancestor.children);
+    });
+  }
```
Caele: This will lead to duplicates in the tree.
in free mode we could just as well keep the topNode static to the root OR if we want options later on just go up a few levels, then start adding nodes (instead of adding normally first, then ancestors.)

haxxmaxx: The upside of having a topNode (which probably should be called main or active now...) is that the only thing that differs between the two modes is ancestors. So we can reuse the logic from before.

niekvanstaveren: We could easily filter out the duplicates here actually. It would be great if we can keep the expandedState the same for all the navigation modes we use.

niekvanstaveren: Should be fixed now

niekvanstaveren:
Added unit test and merged master

PR #66, opened 2020-02-13
fix: setting viewstate in snapshot data

henrikgildeus:
This fixes view state to be correct in snapshots in Storytelling

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -8,6 +8,7 @@ import {
   useTheme,
   useSelections,
   useAction,
+  onTakeSnapshot,
   useOptions,
   useImperativeHandle,
   useConstraints,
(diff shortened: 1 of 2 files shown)


haxxmaxx:
I still get an all white snapshot in story mode

Caele:
> I still get an all white snapshot in story mode

Took a quick look, could be that the layout saved does not include the additional data pages. So might need to add those to the snapshotlayout.

niekvanstaveren:
> > I still get an all white snapshot in story mode
> 
> Took a quick look, could be that the layout saved does not include the additional data pages. So might need to add those to the snapshotlayout.

I think @henrikgildeus saw a difference between his docker environment and when running in desktop/elastic environment. I think he talked to Miralem about it, but not sure what the status is here

henrikgildeus:
Yes, like Niek says. I've tested snapshots and they work fine in Elastic env and Desktop. In local Docker env it behaves very strange and I can't get my head around it. Also, the debugger is lost in the code, so it's hard to say what goes on there. I also checked out some old commits from sn-org-chart master and it was the same. So, this is some separate weird thing

henrikgildeus:
Btw, tested it again in local Docker now. It worked when creating a new org chart and then snapshot it :)

Caele:
> Btw, tested it again in local Docker now. It worked when creating a new org chart and then snapshot it :)

ah nice, good `ol turnin´ it on an´off again`

PR #71, opened 2020-02-17
Tsm/translations

Caele:
Add use of translator. Strings are still subject to check from doc, but that will be done in the client - then updated here.
Also fix for dimensions limits.

### DIFF
--- a/src/extension/meta.json
+++ b/src/extension/meta.json
@@ -1,8 +1,9 @@
 {
-  "name": "Organizational Chart",
-  "description": "Chart to visualise organisations",
+  "name": "Org chart",
+  "description": "Displays the relations of an organization or similarly structured data. Starts at the top and allows for navigation through the hierachy.",
   "icon": "network",
   "preview": "preview.png",
   "author": "QlikTech International AB",
+  "supernova": true,
   "type": "visualization"
 }
--- a/src/locale/en-US.js
+++ b/src/locale/en-US.js
@@ -1,15 +1,16 @@
 // Local translations when used outside of sense
 export default {
   'Object.OrgChart': 'Org chart', // Name of chart, to be added to sense when native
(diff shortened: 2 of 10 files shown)


src/locale/en-US.js
```diff
@@ -0,0 +1,17 @@
+// Local translations when used outside of sense
+export default {
+  'Object.OrgChart': 'Org chart', // Name of chart, to be added to sense when native
+  'Object.OrgChart.Description':
+    'Displays the relations of an organization or similarly structured data. Starts at the top and allows for navigation through the hierachy.', // Description of chart, to be added to sense when native
+  'Object.OrgChart.MaxData': 'The maximum data limit is reached. The tree may display incorrectly.',
+  'Object.OrgChart.DummyRoot': 'Root root',
```
haxxmaxx: why is this needed?

Caele: good question, intended for tooltip on the extra root nodes, but probably not needed. Also strangely written, i'll remove it.

src/extension/meta.json
```diff
@@ -1,8 +1,9 @@
...
-  "name": "Organizational Chart",
-  "description": "Chart to visualise organisations",
+  "name": "Org chart",
+  "description": "Displays the relations of an organization or similarly structured data. Starts at the top and allows for navigation through the hierachy.",
   "icon": "network",
   "preview": "preview.png",
   "author": "QlikTech International AB",
+  "supernova": true,
   "type": "visualization"
 }
```
niekvanstaveren: Would it be possible to have only meta.json or sn-org-chart.qext ?? The files are quite similar. Maybe something for after first release though

Caele: yea, the meta file is actually only for when building with `nebula sense`, that doesn't work right now and as it won't really help us anyway I haven't pursued that. We can mostly ignore that file for now, and likely remove it later.

PR #73, opened 2020-02-17
feat: added zoom and pan for viewstate

niekvanstaveren:
When switching between edit and analysis mode everything will re-render now. Tried to keep the state but that did not really work for zooming/panning. 

We might have another try later, but wanted to get the viewstate in for download/export at least

**EDIT: We do keep the zoom pan state between edit and analysis mode now**

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -185,6 +185,7 @@ export default function supernova(env) {
             selectionsAndTransform,
             selectionState,
             useTransitions: expandedState.useTransitions,
+            element,
           });
         }
       }, [expandedState, objectData, selectionState]);
--- a/src/index.js
+++ b/src/index.js
@@ -48,7 +48,7 @@ export default function supernova(env) {
       const [opts] = useState(options);
       const selectionsAPI = useSelections();
       const constraints = useConstraints();
-      const [selections] = useState({ api: selectionsAPI, setState: setSelectionState, linked: false });
+      const [selections] = useState({ api: selectionsAPI, setState: setSelectionState, linked: false, constraints });
 
       const resetSelections = () => {
         setSelectionState([]);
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -150,7 +152,7 @@ export default function supernova(env) {
               });
           }
         }
-      }, [element, dataTree]);
+      }, [element, dataTree, constraints]);
```
haxxmaxx: since we listen to `constraints` here, we probably don't need to have it in other `useEffects`, since they listen to `objectData`

src/index.js
```diff
@@ -39,6 +39,7 @@ export default function supernova(env) {
       const [expandedState, setExpandedState] = useState(null);
       const [linked, setLinked] = useState(false);
       const [selectionState, setSelectionState] = useState([]);
+      const [transform, setTransform] = useState(null);
```
haxxmaxx: Would this be the same state as Elise has in her PR?

src/index.js
niekvanstaveren: Yes good one!

src/index.js
niekvanstaveren: Yes, Elise will merge first and then I'll see about the conflicts :)

src/index.js
niekvanstaveren: Actually found a way to keep the zoom/panning between edit/analysis mode. Changes the filter function though, so something you might want to have a look at

src/index.js
```diff
@@ -48,7 +48,13 @@ export default function supernova(env) {
       const [opts] = useState(options);
       const selectionsAPI = useSelections();
       const constraints = useConstraints();
-      const [selectionsAndTransform] = useState({ api: selectionsAPI, setState: setSelectionState, linked: false, transform: {} });
+      const [selectionsAndTransform] = useState({
```
niekvanstaveren: Yes we would probably have to rename this one again! Any suggestions are welcome of course

haxxmaxx: selectionAndTransformAndConstraints

haxxmaxx: Can't think of anything on top of my head. Maybe something to indicate that is persisted and stuff can be update updated without reseting the entire object.

PR #76, opened 2020-02-18
fix: tooltip closed on scroll + measure label

eliseeborn:


src/tree/box.js
```diff
@@ -139,6 +139,14 @@ export default function box(
...
         .html('')
         .attr('style', 'visibility: hidden;opacity: 0;');
+    })
+    .on('wheel', () => {
+      clearTimeout(tooltipOpen);
+      tooltipOpen = -1;
+      clearTimeout(tooltipClose);
+      tooltip
+        .html('')
+        .attr('style', 'visibility: hidden;opacity: 0;');
```
haxxmaxx: Break this out since it is used three times

src/tree/box.js
```diff
@@ -113,7 +113,7 @@ export default function box(
       if (allowTooltips && tooltipOpen === -1 && event.buttons === 0) {
         tooltipOpen = setTimeout(() => {
           tooltip
-            .html(`${d.data.attributes.label || d.data.id}<br />${d.data.attributes.subLabel ? `${d.data.attributes.subLabel}<br />` : ''}${d.data.attributes.extraLabel ? `${d.data.attributes.extraLabel}<br />` : ''}${d.data.measure || ''}`)
+            .html(`${d.data.attributes.label || d.data.id}<br />${d.data.attributes.subLabel ? `${d.data.attributes.subLabel}<br />` : ''}${d.data.attributes.extraLabel ? `${d.data.attributes.extraLabel}<br />` : ''}${d.data.measure ? `${cardStyling.measureLabel ? `${cardStyling.measureLabel}: ` : ''}${d.data.measure}` : ''}`)
```
haxxmaxx: This is pretty hard to read 😅 Does it all have to be inside  `html()` ?

src/tree/box.js
Caele: Should consider moving the tooltip html def into its own file, basically allowing you to call tooltip.open(data) and tooltip.close().

You could pass in the tooltip element to those functions, making them akin to stateless.
Also, you can avoid setting .html('') each time, it should not be needed?

eliseeborn: yeah, that's a good idea 👍

src/tree/tooltip.js
```diff
@@ -0,0 +1,57 @@
+import { select } from 'd3';
+import constants from './size-constants';
+
+let tooltipOpen = -1;
```
haxxmaxx: global variables :( I think can get weird with multiple orgcharts in sense

eliseeborn: yeah, that is definitely not working right now :/

Caele: It can be fine, as long as you never need multiple tooltips at the same time. But you also need to use the same element then, else you'll run into issues not having closed a tooltip.

So I think either scope the tooltip timeouts to the box function again, or move the tooltip element onto the html body. That causes some other issues, so best keep it inside the org chart div.

eliseeborn: the problem isn't even the opening and closing, when I have two org charts, the tooltip is over the card in the other orgchart than the one I'm hovering over.
but I guess that would be solveable by going back to passing around the reference to the tooltip

eliseeborn: as well as scoping the timeouts to box I guess

Caele: you could just build it as a class/object of sorts:
- in index you initialize it with the element and save the tooltip reference to the selectionsAndTransform object:
`selectionsAndTransform.tooltip = tooltip.init(element)` -> that can save timouts on itself etc
- Pass that to box.js
- Profit

PR #80, opened 2020-02-18
fix: topnode id not present after selection

niekvanstaveren:


src/tree/render.js
```diff
@@ -4,8 +4,17 @@ import box from './box';
 import createPaths from './path';
 import transform from './transform';
 
-export const filterTree = ({ topId, isExpanded, expandedChildren }, nodeTree) => {
-  const topNode = nodeTree.descendants().find(node => node.data.id === topId) || nodeTree;
+export const filterTree = ({ topId, isExpanded, expandedChildren }, nodeTree, setExpandedState) => {
+  let topNode = nodeTree.descendants().find(node => node.data.id === topId);
+  if (!topNode) {
+    topNode = nodeTree;
+    setExpandedState({
```
haxxmaxx: This will trigger paintTree to run again... Why do we have to do this?

niekvanstaveren: I was thinking we needed to reset the expanded state somewhere when the top id is not in the selected nodes. Couldn't really find a good place to do this, but maybe not even necessary. Let me check

niekvanstaveren:
This is actually enough. We don't reset the navigation state, but should not be necessary

PR #82, opened 2020-02-18
fix: adjust positions

Caele:
Adjusts the positions of the expand buttons and the path, making it a bit clearer to which node the expand button belongs.

Before:
image

After:
image

### DIFF
--- a/src/styles/paths.less
+++ b/src/styles/paths.less
@@ -4,7 +4,7 @@
 
     .sn-org-path {
       fill: none;
-      stroke: #8c8c8c;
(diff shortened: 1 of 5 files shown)


haxxmaxx:
Looks better, but a not super consistent for leaf branches. the "elbow" is closer to the button
Screenshot 2020-02-18 at 17 01 18

Caele:
> Looks better, but a not super consistent for leaf branches. the "elbow" is closer to the button
> Screenshot 2020-02-18 at 17 01 18

Moved them a bit:
image

PR #83, opened 2020-02-18
feat: limit amount of children showing

niekvanstaveren:


niekvanstaveren:
@Caele I agree that 100 is a lot. I think that is the number we agreed upon, but we can always change this of course. The amount of root nodes is actually also limited in this case

src/utils/tree-utils.js
```diff
@@ -175,14 +175,17 @@ export function createNodes(matrix, attributeIndecies, status, navigationMode, t
...
     const parentNode = nodeMap[node.parentId];
     node.parent = parentNode;
     if (parentNode) {
-      parentNode.children.push({ childNumber: parentNode.children.length, ...node });
+      parentNode.children.length > 100
+        ? (maxNodeWarning = true)
+        : parentNode.children.push({ childNumber: parentNode.children.length, ...node });
     } else {
-      rootNodes.push(node);
+      rootNodes.length > 100 ? (maxNodeWarning = true) : rootNodes.push(node);
```
niekvanstaveren: This should limit amount of rootnodes

PR #84, opened 2020-02-18
fix: Update styling when theme is changed

haxxmaxx:
Should we still set `layout` in `usePromise`? Feels like a bit of a stretch to redo everything when `layout` has not change

### DIFF
--- a/src/utils/styling.js
+++ b/src/utils/styling.js
@@ -1,13 +1,13 @@
 import colorUtils from './color-utils';
 
-export function getColor(reference, palette, defaultColor) {
+export function getColor(reference, Theme, defaultColor) {
   let color;
   switch (reference.colorType) {
     case 'byExpression':
       color = colorUtils.resolveExpression(reference.colorExpression);
       break;
     case 'colorPicker':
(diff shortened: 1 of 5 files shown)


Caele:
> Should we still set `layout` in `usePromise`? Feels like a bit of a stretch to redo everything when `layout` has not change

I guess you could divide them up to get slightly better performance when changing themes. I don't think it matters much as we shouldn't need to fetch the data again.

src/utils/styling.js
```diff
@@ -1,13 +1,13 @@
...
 
-export function getColor(reference, palette, defaultColor) {
+export function getColor(reference, Theme, defaultColor) {
   let color;
   switch (reference.colorType) {
     case 'byExpression':
       color = colorUtils.resolveExpression(reference.colorExpression);
       break;
     case 'colorPicker':
-      color = colorUtils.resolveColor(reference.color, palette);
```
niekvanstaveren: I think this is the only place we use this function. We could remove it from utils then

haxxmaxx: ah true, I think Miralem removed it from the action button.

PR #85, opened 2020-02-18
fix: add label and string for linked selections

Caele:


niekvanstaveren:
Was working on the same including some other changes for the icon, but this look perfect 👍

PR #86, opened 2020-02-19
fix: update strings and add description note

Caele:


src/extension/ext-raw.js
```diff
@@ -73,6 +74,11 @@ export default {
                   },
                 ],
               },
+              desc: {
+                component: 'text',
+                translation: 'Object.OrgChart.ExtraLabelDesc',
+                style: 'hint',
+              },
```
Caele: had to place it after all the attributes as that component doesn't allow for non-attribute things. And adjusting that in the client looked a bit risky.

PR #88, opened 2020-02-19
fix: add data to snapshot

Caele:
The snapshot now only stores the data it needs, which is the nodes you see, plus their children (to get the numbers right, bit of a shortcut).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -13,7 +13,6 @@ import {
   useImperativeHandle,
   useConstraints,
   useTranslator,
-  useRect,
 } from '@nebula.js/supernova';
 import properties from './object-properties';
 import data from './data';
@@ -58,8 +57,6 @@ export default function supernova(env) {
         transform: {},
         constraints,
       });
-
-      const rect = useRect();
 
       const translator = useTranslator();
       useEffect(() => {
@@ -216,10 +213,9 @@ export default function supernova(env) {
             setTransform,
             transformState: (viewState && viewState.transform) || {},
             selectionsAndTransform,
-            rect,
           });
         }
-      }, [objectData, rect]);
+      }, [objectData]);
 
       const createViewState = () => {
         const vs = {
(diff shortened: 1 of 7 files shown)


src/tree/render.js
```diff
@@ -4,17 +4,24 @@ import box from './box';
...
-      if (child.children && expandedChildren.indexOf(child.data.id) !== -1) {
-        child.children.forEach(grandChild => {
-          subTree.push(grandChild);
-        });
+      if (child.children) {
+        if (expandedChildren.indexOf(child.data.id) !== -1 || extended) {
+          child.children.forEach(grandChild => {
+            subTree.push(grandChild);
+            if (expandedChildren.indexOf(child.data.id) !== -1 && extended && grandChild.children) {
+              child.children.forEach(extendedChild => {
```
haxxmaxx: Doesn't this just att the granchild twice?

niekvanstaveren: Yes this should be grandChild.children.forEach I think

Caele: yea it should, how the heck did this work then, or seem to work.

Caele: I did not look closely at that =)

src/index.js
```diff
@@ -155,7 +154,7 @@ export default function supernova(env) {
        * - render [allNodes, positioning, styling, rect]
        */
 
-      usePromise(() => {
+      const [dataTree] = usePromise(() => {
```
Caele: This was likely not needed, but removes a state for us.

src/utils/tree-utils.js
```diff
@@ -181,11 +186,11 @@ export function createNodes(matrix, attributeIndecies, status, navigationMode, t
     const parentNode = nodeMap[node.parentId];
     node.parent = parentNode;
     if (parentNode) {
-      parentNode.children.length > 100
+      parentNode.children.length > 98
```
Caele: Changed this to 98, so that the max number of children becomes 99, which fits in the +/- box.

PR #89, opened 2020-02-19
fix: tooltip close on scroll and click

niekvanstaveren:
@Caele @eliseeborn This seems to work for me. Am I missing something here?

### DIFF
--- a/src/tree/tooltip.js
+++ b/src/tree/tooltip.js
@@ -31,13 +31,17 @@ export function openTooltip(tooltip, d, containerHeight, cardStyling, x, y, sel)
   const measure = d.data.measure
     ? `${cardStyling.measureLabel ? `${cardStyling.measureLabel}: ` : ''}${d.data.measure}`
     : '';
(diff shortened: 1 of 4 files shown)


niekvanstaveren:
Ah I am missing something :)

src/tree/tooltip.js
```diff
@@ -0,0 +1,47 @@
...
+    ? `${cardStyling.measureLabel ? `${cardStyling.measureLabel}: ` : ''}${d.data.measure}`
+    : '';
+  tooltip.active = true;
+  setTimeout(() => {
+    if (tooltip.active) {
+      tooltip
+        .html(`${label}<br />${subLabel}${extraLabel}${measure}`)
+        .attr('style', () => getTooltipStyle(d, containerHeight, x, y, sel));
+    }
+  }, 250);
```
Caele: Moving your mouse quickly will queue lots of timouts, so do a check for tooltip.active and cancel any current timeout (store the timeout on tooltip).

PR #90, opened 2020-02-19
test: cover util files

haxxmaxx:
Pausing here so this wont get too big.
Todo in next PR:
- `tooltips`
- `transform`
- the async stuff in `tree-utils`
- some cases in `position`

`render` and `index` are so much d3 and nebula so they are overkill to mock IMO.

### DIFF
--- a/src/utils/__tests__/selections.spec.js
+++ b/src/utils/__tests__/selections.spec.js
@@ -1,0 +1,66 @@
+import selections from '../selections';
+import defaultValues from '../../__tests__/default-orgchart-props';
+
+
+describe('selections', () => {
+  describe('select', () => {
+    const { select } = selections;
+    let node;
+    let selectionObj;
+    let selectionState;
+    let isActive;
+    beforeEach(() => {
+      node = JSON.parse(JSON.stringify(defaultValues.nodes));
+      isActive = false;
+      selectionObj = {
+        api: {
+          isActive: () => isActive,
+          clear: sinon.spy(),
+          select: sinon.spy(),
+          begin: sinon.spy(),
+        },
+        linked: false,
+        setState: sinon.spy(),
+      };
+      selectionState = [];
+    });
+    it('should early return if elemNo is negative', () => {
+      node.data.elemNo = -2;
+      select(node, selectionObj, selectionState);
+      expect(selectionObj.setState).to.not.have.been.called;
+    });
+    it('should early return if elemNo is negative', () => {
+      node.data.isLocked = true;
+      select(node, selectionObj, selectionState);
+      expect(selectionObj.setState).to.not.have.been.called;
+    });
+    it('should call begin sand setState to node id', () => {
+      select(node, selectionObj, selectionState);
+      expect(selectionObj.api.begin).to.have.been.calledOnce;
+      expect(selectionObj.api.select).to.have.been.calledOnce;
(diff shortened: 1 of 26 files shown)


niekvanstaveren:
Doing a great job here @haxxmaxx, but can we have it in multiple PR's so it will be easier to check?

src/utils/selections.js
```diff
@@ -10,7 +10,7 @@ export default {
       if (node.data.isLocked) {
         return;
       }
-      if (!selections.api.isActive() || !selectionState) {
+      if (!selections.api.isActive() || !selectionState) { // selectioState !== [] ?
```
haxxmaxx: @Caele Are we suppose to check if selectionState is emtpy?

src/utils/__tests__/selections.spec.js
```diff
@@ -0,0 +1,73 @@
...
+        setState: sinon.spy(),
+      };
+      selectionState = [];
+    });
+    it('should early return if elemNo is negative', () => {
+      node.data.elemNo = -2;
+      select(node, selectionObj, selectionState);
+      expect(selectionObj.setState).to.not.have.been.called;
+    });
+    it('should early return if elemNo is negative', () => {
```
niekvanstaveren: Different test case

src/utils/__tests__/selections.spec.js
```diff
@@ -0,0 +1,73 @@
...
+      node.data.isLocked = true;
+      select(node, selectionObj, selectionState);
+      expect(selectionObj.setState).to.not.have.been.called;
+    });
+    it('should call begin sand setState to node id', () => {
```
niekvanstaveren: Typo

src/utils/__tests__/tree-utils.spec.js
```diff
@@ -18,6 +18,16 @@ describe('tree-utils', () => {
       const node = createNodes(matrix, [], null, null, translator);
       expect(node.id).to.equal('007');
     });
+    it('should add a warning for max data', () => {
+      const matrix = [[{ qText: '007', qElemNumber: 0 }, { qText: '-' }]];
+      const node = createNodes(matrix, [], 'max-data-limit', null, translator);
+      expect(node.warn).to.eql(['Object.OrgChart.MaxData']);
```
niekvanstaveren: Does this function work?

haxxmaxx: equal is equivalent to `===`, which would return false, since it is not the same instance of the array. So `eql` checks the contents instead, kind of... https://stackoverflow.com/questions/36798993/what-is-the-difference-between-equal-and-eql-in-chai-library

niekvanstaveren: Ah nice, I have been using deep equal I think :)

haxxmaxx:
> This will have some merge conflicts with the changes on the zooming though

Yeah I think that's good. After this coverage push, we should try to update the tests to our own changes 😉

niekvanstaveren:
> > This will have some merge conflicts with the changes on the zooming though
> 
> Yeah I think that's good. After this coverage push, we should try to update the tests to our own changes 😉

Hard to disagree with that

src/utils/__tests__/tree-utils.spec.js
haxxmaxx: It is very similar, didn't really understand the difference when I read about it

PR #91, opened 2020-02-20
chore: store dist

Caele:


Caele:
> I assume this will store the extension every time we commit to master?

yea, that is the goal

PR #94, opened 2020-02-20
feat: auto width tooltip and restyle

Caele:
- Updated the styles to better fit the client tooltips
- Switched to auto width, using a dummy parent element to center it
- using width: fit-content to allow overflow (setting a fixed size in IE).

### DIFF
--- a/src/styles/tooltip.less
+++ b/src/styles/tooltip.less
@@ -2,37 +2,58 @@
   .sn-org-tooltip {
     visibility: hidden;
     opacity: 0;
+    /* Unsupported in IE, so there we set width explicitly */
+    width: fit-content;
+
+    @media all and (-ms-high-contrast: none), (-ms-high-contrast: active) {
+      width: 230px;
+    }
+
+    /* Position the tooltip text */
+    position: absolute;
+    z-index: 1;
+  }
+
+  /* Tooltip arrow */
+  .sn-org-tooltip::after {
+    content: '';
+    position: absolute;
+    top: 100%;
+    left: 0%;
+    margin-left: -10px;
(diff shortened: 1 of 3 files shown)


eliseeborn:
it seems to be a little weird on the very right of the chart window
Screenshot 2020-02-20 at 16 26 02

Caele:
> it seems to be a little weird on the very right of the chart window
> Screenshot 2020-02-20 at 16 26 02

yes, i'm using width: fit-content; to fix that, but it doesn't work in IE, so I added a check for that. Still doesn't work in Edge unfortunately. Which browser did you use?

Caele:
See that it didn't like that in firefox either, so might need to fallback to a fixed width anyway. Unless we do some more advanced calculations.

src/styles/tooltip.less
```diff
@@ -2,37 +2,56 @@
...
-    padding: 5px 0;
-    border-radius: 6px;
-
-    overflow-wrap: break-word;
-    word-wrap: break-word;
-    word-break: break-word;
+    /* Unsupported in IE, so there we set width explicitly */
+    width: 230px;
+    width: -moz-max-content;
+    width: fit-content;
```
Caele: Unless the width is set to something, it will not allow the tooltip to expand beyond the chart. This would be solved by moving the element to the body, but we'll leave that for later. In chrome/safari you can set fit-content and it behaves nicely, in firefox -moz-max-content does almost the same thing. In IE I used a fixed width instead, which isn't nice but fine.

eliseeborn: I was using firefox, looks good now :)

PR #95, opened 2020-02-20
feat: correct zoom on first render

niekvanstaveren:
So I looked a bit more into our zooming (mostly because we don't account for the height at the moment). This would also make it possible to get a Home button in by re-rendering everything. What happens:

1. We get new objectData
2. Based on expanded state we filter the tree to get the nodes we want to render
3. bBox of these nodes is determined
4. Calculations for position of the nodes and initial zooming
5. Replace positioning on objectData
6. Setting a new expanded state (if we did not have any)
7. Paint the tree (also done when setting expanded state)

It could definitely use a brush-up before we even think of merging this, but let me know what you think about it. 

Also I haven't been able to get the chart to render in the center of the element. So it will now locate in the top left corner from the start

### DIFF
--- a/src/tree/transform.js
+++ b/src/tree/transform.js
@@ -1,7 +1,7 @@
 import { select, zoom, event, zoomIdentity } from 'd3';
 import constants from './size-constants';
 
-const getBBoxOfNodes = nodes => {
+export const getBBoxOfNodes = nodes => {
   const { cardWidth, cardHeight, buttonHeight, buttonMargin } = constants;
   const bbox = {
     left: Infinity,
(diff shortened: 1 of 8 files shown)


haxxmaxx:
> Also I haven't been able to get the chart to render in the center of the element. So it will now locate in the top left corner from the start

Haven't looked too close, but what about setting the `xActual` to half of `element.clientWidth` if it the parent has no `xActual` like you did before?

niekvanstaveren:
> > Also I haven't been able to get the chart to render in the center of the element. So it will now locate in the top left corner from the start
> 
> Haven't looked too close, but what about setting the `xActual` to half of `element.clientWidth` if it the parent has no `xActual` like you did before?

When you have children expanded you can end up in a situation where the top node should not be in the middle of the element. That is however not the positioning where I face the difficulty. The problem is:

xAxis when zooming in y direction
yAxis when zooming in x direction

niekvanstaveren:
> Think it looks good, hard to follow exactly.
> But I did see some graphical glitches after selections which I could not see on master. Basically after a selection you'd end up with paths lingering from last time.

Actually I saw some problems on the Friday afternoon as well. I don't think it is lingering paths, but rather paths not calculated correctly. I only saw this when making selections with multiple root nodes. Will look at this today (hopefully with some help from @haxxmaxx )

src/tree/transform.js
```diff
@@ -23,6 +23,30 @@ export const getBBoxOfNodes = nodes => {
   };
 };
 
+export const getInitialZoomState = (bBox, element) => {
```
haxxmaxx: This can be tested

niekvanstaveren: Will add some test now

PR #97, opened 2020-02-21
feat: resize for snapshot

niekvanstaveren:


Caele:
Tried a bit, it is not exact but could have to do with changes to the ratio. Sometimes the storytelling view is not exactly the same as the one on the sheet and requires some tinkering.

niekvanstaveren:
> Tried a bit, it is not exact but could have to do with changes to the ratio. Sometimes the storytelling view is not exactly the same as the one on the sheet and requires some tinkering.

niekvanstaveren:
@Caele I'm not sure how we can fix those differences. We check for the axis we need to zoom on and the other axis might show a little bit more, or did you see anything else?

PR #101, opened 2020-02-24
Tsm/touchmode

Caele:
Fixes tooltips on touch:
- Checks the touch-on class that the client uses
- if on, displays tooltips one tap/click. Disables them from mouseover.

### DIFF
--- a/src/tree/box.js
+++ b/src/tree/box.js
@@ -70,17 +70,15 @@ export default function box(
   element,
   tooltip
 ) {
-  const {
-    cardWidth,
-    cardHeight,
-    buttonWidth,
-    buttonHeight,
-    cardPadding,
-    rootDiameter,
-  } = constants;
+  const { cardWidth, cardHeight, buttonWidth, buttonHeight, cardPadding, rootDiameter } = constants;
   const { topId, isExpanded } = expandedState;
   const topNode = nodes.find(node => node.data.id === topId);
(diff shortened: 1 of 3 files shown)


src/tree/box.js
```diff
@@ -70,18 +70,12 @@ export default function box(
...
-    buttonHeight,
-    cardPadding,
-    rootDiameter,
-  } = constants;
+  const { cardWidth, cardHeight, buttonWidth, buttonHeight, cardPadding, rootDiameter } = constants;
   const { topId, isExpanded } = expandedState;
   const topNode = nodes.find(node => node.data.id === topId);
   const ancestorIds = topNode && topNode.parent ? topNode.parent.ancestors().map(anc => anc.data.id) : [];
 
+  const touchmode = document.getElementsByTagName('html')[0].classList.contains('touch-on');
```
Caele: This should likely be replaces by a nebula api or similar.

src/tree/transform.js
```diff
@@ -72,6 +73,7 @@ export function setZooming({ objectData, setTransform, transformState, selection
   const zoomed = () => {
     setTransform({ zoom: event.transform.k / scaleFactor, x: event.transform.x, y: event.transform.y });
     bubbleEvent();
+    closeTooltip(tooltip);
```
haxxmaxx: Is this to close when flipping between landscape/portrait?

Caele: nah it's to close it when swiping. The touchmove listener didn't work from what I could tell.

PR #109, opened 2020-02-26
feat: home button draft

niekvanstaveren:
Had a quick look at a home button. Probably wait with this until next milestone, but it shouldn't be too hard. We should refactor our index.js file a bit I think. We could probably move out a lot of things to make it clearer what is going on

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -29,6 +29,7 @@ import './styles/tooltip.less';
 import './styles/paths.less';
 import './styles/warnings.less';
 import './styles/nodes.less';
+import './styles/homebutton.less';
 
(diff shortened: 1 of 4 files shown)


Caele:
> Had a quick look at a home button. Probably wait with this until next milestone, but it shouldn't be too hard. We should refactor our index.js file a bit I think. We could probably move out a lot of things to make it clearer what is going on

yea, probably the first thing we should do. Just to clean things up. I know CAM mentioned that you would be able to split things up better after a nebula update. I looked a bit at the piechart and that can give us a nice guidance on how to do it.

PR #111, opened 2020-02-26
test: unit test tree utils

niekvanstaveren:


src/utils/__tests__/tree-utils.spec.js
```diff
@@ -131,4 +195,44 @@ describe('tree-utils', () => {
...
+        qHyperCube: {
+          qDimensionInfo: [{}, {}],
+          qDataPages: [{ qArea: { qHeight: 1 }, qMatrix: [] }],
+          qSize: { qcy: 1 },
+        },
+      };
+      const result = await transform({ layout, model, translator });
+      expect(result).to.be.null;
+    });
+    it('should return null for empty dataMatrix', async () => {
```
haxxmaxx: description

src/utils/__tests__/tree-utils.spec.js
```diff
@@ -12,6 +21,65 @@ function generateMatrix(numRows, childCount = 1) {
...
+      const result = await fetchPage([], [], model, 1, 1, 1, 10);
+      expect(result).to.equal('');
+    });
+    it('should return max data when limit reached', async () => {
+      const result = await fetchPage([], [], model, 10, 1, 1, 3);
+      expect(result).to.equal('max-data-limit');
+    });
+  });
+
+  describe('getDataMatrix', () => {
```
haxxmaxx: One with empty datapages would be nice

niekvanstaveren: Which part would be empty then?

niekvanstaveren: Like this https://github.com/qlik-oss/sn-org-chart/pull/111/commits/dd8e280af7442700a736eb7b9817dbf183ba7107 ?

haxxmaxx: I looked at coveralls and one case on line 97 was missed. Looking at the code, I realize it is the opposite. dataPages is never longer than 1, so i is never greater than 0.

niekvanstaveren: I'm not even sure that is possible coming back from engine, but added a test for it anyway (since we do have the logic for some reason)

PR #114, opened 2020-03-03
refactor: index

haxxmaxx:
- break out snapshot stuff to separate file + mov0e `createSnapshotData` there
- break out selection stuff
- split up `selectionAndTransform`


src/index.js
```diff
@@ -6,24 +6,21 @@ import {
...
   useTranslator,
 } from '@nebula.js/supernova';
 import properties from './object-properties';
 import data from './data';
 import ext from './extension/ext';
-import { paintTree, preRenderTree, createSnapshotData, filterTree } from './tree/render';
-import position from './tree/position';
+import snapshot from './snapshot';
+import selectionHandler from './selections-handler';
+import { paintTree, preRenderTree } from './tree/render';
```
niekvanstaveren: We should come up with a better name for preRenderTree I think

src/index.js
```diff
@@ -37,248 +34,105 @@ export default function supernova(env) {
...
-      const [opts] = useState(options);
       const rect = useRect();
-      const selectionsAPI = useSelections();
       const constraints = useConstraints();
-      const [selectionsAndTransform] = useState({
-        api: selectionsAPI,
-        setState: setSelectionState,
-        linked: false,
+      const translator = useTranslator();
+      const [preRenderData, setPreRenderData] = useState(null);
```
niekvanstaveren: So what is preRenderData?

src/index.js
```diff
@@ -37,248 +34,105 @@ export default function supernova(env) {
...
-      }, [translator]);
-
-      const resetSelections = () => {
-        setSelectionState([]);
-      };
-      const resetSelectionsAndLinked = () => {
-        setLinked(false);
-        setSelectionState([]);
-      };
+      }, [translator.language()]);
```
niekvanstaveren: Did you find that updating the language doesn't work?

src/index.js
```diff
@@ -37,248 +34,105 @@ export default function supernova(env) {
...
-        api: selectionsAPI,
-        setState: setSelectionState,
-        linked: false,
+      const translator = useTranslator();
+      const [preRenderData, setPreRenderData] = useState(null);
+      const [styling, setStyling] = useState(null);
+      const [expandedState, setExpandedState] = useState(null);
+      const [transform, setTransform] = useState(null);
+      const [initialZoom, setInitialZoom] = useState(null);
+      const [storageState] = useState({
```
niekvanstaveren: What is the storageState?

src/index.js
```diff
@@ -37,248 +34,105 @@ export default function supernova(env) {
...
+            setExpandedState,
+          });
           if (preRender) {
-            setObjectData(preRender);
+            setPreRenderData(preRender);
           }
         }
       }, [element, dataTree, constraints]);
 
+      // Updates snapshot??
```
niekvanstaveren: Yes updates the snapshot when resizing

src/card/card.js
```diff
@@ -16,9 +16,9 @@ export function getFontColor(cardStyling, backgroundColor) {
   return cardStyling.fontColor;
 }
 
-export default (data, cardStyling, sel, selectionState) => {
-  const selections = sel.api;
-  const isSelected = selections && selections.isActive() && selectionState.indexOf(data.elemNo) !== -1;
+export default (data, cardStyling, selectionObj) => {
+  const { api, state } = selectionObj;
+  const isSelected = api && api.isActive() && state.indexOf(data.elemNo) !== -1;
```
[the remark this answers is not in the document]
haxxmaxx: Well, sure. I mean this is the only api we are using more or less at this point? In that case we should rename `state` as well

src/index.js
haxxmaxx: What `preRenderTree` returns. `objectData` did not make any real sense to me, at least this refers to where it comes from... But yeah maybe a better name for preRender then rename this accordingly.

src/index.js
haxxmaxx: `selectionsAndTransorm` minus the selection stuff. So constraints and transform at this point. Wanted to call it something with state, since it is a state. Name suggestions are very welcome 🙃

src/index.js
haxxmaxx: Yes, it works similar to how constraints work. Miralem told me this a while back, came to mind now.

haxxmaxx:
> looks ok I think, getting some stuff out will help with continuing this.
> We should also look at how the preRender things are done, basically we should only create the container elements once, not when data changes as we do now. Then in a separate pass attach listeners/clear children or whatever else needed.

I can look at that in the next PR, this one is already pretty big. Will just push some name changes first :)

niekvanstaveren:
> looks ok I think, getting some stuff out will help with continuing this.
> We should also look at how the preRender things are done, basically we should only create the container elements once, not when data changes as we do now. Then in a separate pass attach listeners/clear children or whatever else needed.

So we clear everything now because of the zooming. The zoomwrapper holds the state of the zooming which we want to clear if we re-render everything. So far I have not come up with something good to replace this. Touche works quite well for zooming on touch devices, but then we still need to zoom on desktop as well of course.

src/index.js
niekvanstaveren: I think the language only gets updated in Sense after a full refresh right?

src/index.js
```diff
@@ -132,7 +132,7 @@ export default function supernova(env) {
         }
       }, [expandedState, containerData, selectionObj.state]);
 
-      snapshot(expandedState, containerData, layout, transform, initialZoom);
+      snapshot(expandedState, containerData, layout, transform, initialZoom, element);
```
niekvanstaveren: Why are we passing it in here. In the function we have useElement and don't get it from the parameters

haxxmaxx: SHould not be there good catch

PR #115, opened 2020-03-05
fix: split touch and scroll zooming

niekvanstaveren:
So this is how far I got now. I'm splitting up the touch zooming and the scroll zooming. For scroll zooming D3 zooming is still used. I tried to see if we could have some other way, but ran into all kinds of issues. Please have a look and let me know what you think.

Still to do:

- [x] Constraints for touch zooming
- [x] Verifying export and snapshots
- [x] Fix unit tests for adding touchejs library


src/utils/interaction-utils.js
```diff
@@ -0,0 +1,7 @@
+class Interactions {
+  constructor() {
+    this.swiping = false;
+  }
+}
+
+export default Interactions;
```
niekvanstaveren: I was thinking of having more interactions in here, now it seems a bit of a lonely class. Will probably have to do this in a different way

src/tree/box.js
```diff
@@ -101,6 +102,16 @@ export default function box(
...
     })
+    .each((node, index, cards) => {
+      Touche(cards[index]).tap({
+        end: () => {
+          if (!selectionsAndTransform.constraints.active && node.data.id !== 'Root') {
+            touchmode && openTooltip(tooltip, node, element.clientHeight, cardStyling, x, y, selectionsAndTransform, 0);
+            selections.select(node, selectionsAndTransform, selectionState);
+          }
+        },
+      });
```
Caele: This call returns an unbind function, make sure to run those to remove listeners. I think doing this for each node is ok, as touch only registers a single touchevent on window, but best keep an eye on it.

src/tree/transform.js
```diff
@@ -116,15 +168,15 @@ export function setZooming({ objectData, setTransform, transformState, selection
         () =>
           !selectionsAndTransform.constraints.active &&
           event.type !== 'dblclick' &&
-          !(event.type === 'mousedown' && event.which === 3)
+          !(event.type === 'mousedown' && event.which === 3) &&
+          event.type !== 'touchstart'
```
Caele: Should be able to only check for event.type === "mousewheel" or something.

This should be ok to keep, saves us some trouble with implementing the correct scroll/pos logic.

niekvanstaveren: I might need to rename this. This is also panning for desktop. For D3 these need to be combined otherwise x and y are not in sync

src/tree/box.js
niekvanstaveren: The Touche() call? Where can I find that?

Caele: hmm, maybe it doesn't. just looked like it in the client, but that one wasn't used.  You can do:
`Touche(item.container).off('*', 'namespace');`
for a:
```
Touche(cards[index]).tap({
    id: 'namespace',
    end: () => {
       ...
       }
     },
   });
```

niekvanstaveren: Ah that should work yes. When do we want to clean them up? I guess the only place would be then in preRenderTree before removing the whole innerHTML, but wouldn't  the listeners be removed then anyway?

Caele: I guess they would yea, but best the thorough on these things. Suddenly we stop removing the elements and are stuck with multiple listeners.

niekvanstaveren: Had to create a class for the elements to be removed. You need to have the specific element to remove the listener from it

src/tree/box.js
```diff
@@ -101,6 +101,18 @@ export default function box({
         selections.select(node, selectionObj);
       }
     })
+    .each((node, index, cards) => {
+      if (!interactions.isIE) {
+        Touche(cards[index]).tap({
+          end: () => {
+            if (!wrapperState.constraints.active && node.data.id !== 'Root') {
```
haxxmaxx: Pretty sure we don't have to check for `Root` anymore, since it is an empty div I think

src/tree/transform.js
```diff
@@ -66,10 +67,10 @@ export const getTranslations = (bBox, height, width) => {
 };
 
 export const applyTransform = (eventTransform, svg, divBox, width, height) => {
-  const scaleFactor = eventTransform.k;
+  const scaleFactor = eventTransform.zoom;
```
haxxmaxx: Could deconstruct here

src/tree/transform.js
```diff
@@ -79,33 +80,116 @@ export const applyTransform = (eventTransform, svg, divBox, width, height) => {
...
+  const translate = (e, data, saveState) => {
+    if (!wrapperState.constraints.active) {
+      closeTooltip(tooltip);
+      const { deltaX, deltaY } = data.swipe;
+      const newX = wrapperState.transform.x + deltaX;
+      const newY = wrapperState.transform.y + deltaY;
+      applyTransform({ x: newX, y: newY, zoom: wrapperState.transform.zoom }, svg, divBox, width, height);
+      if (saveState) {
+        setTransform({ x: newX, y: newY, zoom: wrapperState.transform.zoom });
+        setTimeout(() => {
```
haxxmaxx: why is the timeout needed?

src/tree/transform.js
```diff
@@ -79,33 +80,116 @@ export const applyTransform = (eventTransform, svg, divBox, width, height) => {
...
-    bubbleEvent();
+  const pinchZoom = (e, data, saveState) => {
+    if (!wrapperState.constraints.active) {
+      closeTooltip(tooltip);
+      const oldZoom = wrapperState.transform.zoom;
+      const newZoom = Math.max(Math.min(oldZoom / (1 / data.scale), maxZoom), minZoom);
+      if (newZoom !== oldZoom) {
+        const zoomDelta = newZoom / oldZoom;
+        const newX = wrapperState.transform.x * zoomDelta + (width - zoomDelta * width) / 2;
+        const newY = wrapperState.transform.y * zoomDelta + (height - zoomDelta * height) / 2;
```
haxxmaxx: A comment or two wouldn't hurt

src/tree/transform.js
niekvanstaveren: The problem here is that zoom is already defined as an import function

haxxmaxx: ok, could not deconstruct here then 🙃

src/tree/transform.js
niekvanstaveren: When you stop swiping or zooming there is a touch tap event triggered (the end event). We don't want to trigger the selection state at the end of zooming/panning. If you remove the timeout you the swiping state would be false again when you reach the touch tap end event

Caele: Need to keep an eye on this. There should be a better way, but I know it is generally very problematic at times. Could you move the timeout and swiping = false to the end() handler instead? just to make things clearer? keeping it in translate kind of hide it

niekvanstaveren: Yeah that should be no problem

PR #119, opened 2020-03-09
fix: position with multiple root nodes

niekvanstaveren:


niekvanstaveren:
> Just fix the test, create that it failed tho :)

Yeah I'm not sure why I put that offset in there in the first place, but good that the UT triggered on this :)

PR #121, opened 2020-03-10
DO NOT MERGE! feat: first draft images

niekvanstaveren:
So I think the implementation in the org chart should not be too difficult. I have showed Christofer how it looks with an image and he will come up with a design for the cards containing an image.

image

### DIFF
--- a/src/card/card.js
+++ b/src/card/card.js
@@ -23,7 +23,11 @@ export default (data, cardStyling, selectionObj) => {
   const topColor = colorUtils.getDarkColor(backgroundColor);
   const fontColor = getFontColor(cardStyling, backgroundColor);
   const attributes = data.attributes || {};
-  let html = `<div class="sn-org-card-title">${attributes.label || data.id}</div>`;
+  let html = '<div class="sn-org-card-content">';
+  if (attributes.imageURL) {
+    html += `<img class="sn-org-card-image" src="${attributes.imageURL}"/>`;
+  }
+  html += `<div class="sn-org-card-textarea"><div class="sn-org-card-title">${attributes.label || data.id}</div>`;
   if (attributes.subLabel) {
     html += `<div class="sn-org-card-label">${attributes.subLabel}</div>`;
   }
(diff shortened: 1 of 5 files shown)


haxxmaxx:
ok I'll go ahead and merge this then

src/card/card.js
```diff
@@ -23,7 +23,11 @@ export default (data, cardStyling, selectionObj) => {
   const topColor = colorUtils.getDarkColor(backgroundColor);
   const fontColor = getFontColor(cardStyling, backgroundColor);
   const attributes = data.attributes || {};
-  let html = `<div class="sn-org-card-title">${attributes.label || data.id}</div>`;
+  let html = '<div class="sn-org-card-content">';
+  if (attributes.imageURL) {
+    html += `<img class="sn-org-card-image" src="${attributes.imageURL}"/>`;
```
Caele: think the URL needs to be validated so there is no possibility for script injection.

src/card/card.js
```diff
@@ -23,7 +23,11 @@ export default (data, cardStyling, selectionObj) => {
...
+  }
+  html += `<div class="sn-org-card-textarea"><div class="sn-org-card-title">${attributes.label || data.id}</div>`;
```
Caele: hmm, possible that these also should be escaped?
`='<script type="text/javascript">function() {alert.show("hey")}()</script>'`
Doesn't really do anything in the client (chrome) but feels a bit risky as I can see it inject the node.

src/card/card.js
niekvanstaveren: Well that would be a big check I think. Also everyone can see which checks we do, but more important which checks we don't do

src/card/card.js
Caele: can't get it to run any of my injected code, so might be fine. But still should probable add a html encoding to all of these.

src/card/card.js
Caele: hmm, I was thinking a encode URL would do it, but googling doesn't give me any of that. 
Think we can look at how it is done in the T&I object, `web/assets/client/services/media-tool/formatters/link.js`

PR #126, opened 2020-03-18
feat: chained select as default

niekvanstaveren:
This will set chained selections as default. It will need a new icon for the selection bar, but you can already have a look at the PR


src/utils/selections.js
```diff
@@ -22,13 +22,14 @@ export default {
...
-      if (linked) {
+      if (!singleSelect) {
         linkedIds = getAllTreeElemNo(node, activate);
       }
       if (!activate) {
         newState.splice(ind, 1);
         linkedIds.forEach(id => {
-          newState.splice(newState.indexOf(id), 1);
+          const idInd = newState.indexOf(id);
+          idInd !== -1 && newState.splice(idInd, 1);
```
niekvanstaveren: This is actually to fix a bug that Mirja found. When deselecting from chained selection we removed other selections in the array

haxxmaxx:
Looks good. If I understand AFAS requirements correctly, a selection outside of the org chart, should be a chain selection. Will that be solved in a different PR?

niekvanstaveren:
> Looks good. If I understand AFAS requirements correctly, a selection outside of the org chart, should be a chain selection. Will that be solved in a different PR?

I don't think we can make that change happening actually. Then the selections you make would all of a sudden change when you have an org chart in the grid? 

Apparently the chained selection was hard to find for a lot of people and the assumption is that chained selection would be the default. So this takes care of that. We had a short discussion on a search box to find the node you are looking for. There we could possibly also add something with selections. But that is not in scope yet.

haxxmaxx:
Yeah that makes sense, would be really weird to brake the pattern we have now for selections in the client. I guess you have to have some other field that you can do selections on if you want that behaviour, as a work around.

PR #127, opened 2020-03-19
feat: expand full tree

niekvanstaveren:
Full tree mode for org chart

TODO

- [x] Snapshot
- [x] Export/Download


src/extension/ext-raw.js
```diff
@@ -122,22 +120,21 @@ export default {
             translation: 'properties.presentation',
             type: 'items',
             items: {
+              navigation: {
+                type: 'items',
+                items: {
+                  navigtaionMode: {
```
Caele: spelling

PR #131, opened 2020-03-24
feat: resize on expand collapse

niekvanstaveren:


src/index.js
```diff
@@ -121,19 +124,22 @@ export default function supernova(env) {
...
+              styling,
+              setExpandedCallback,
+              wrapperState,
+              selectionObj,
+              useTransitions: expandedState.useTransitions,
+              element,
+            });
+          }
+        },
+        layout.resizeOnExpand ? [containerData, selectionObj.state] : [expandedState, containerData, selectionObj.state]
```
Caele: hmm, not sure you are allowed to do conditionals like this for hooks. Can't you just do this check in the "if" surrounding the paintTree call instead?

niekvanstaveren: It seems to work, but I'll convert since it's not very clear to read in this way

Caele: I think it depends on the internal implementation of hooks, for example if the dependencies are updated on each useEffect call or not. You could achieve the same thing with a flagged state, that only updates if resizeOnExpand is true, but still, clearer if you just check it inside the hook. Performance wise it should just be a single extra function call each loop.

niekvanstaveren: Moved expanded state into its own useEffect cycle

PR #139, opened 2020-04-20
fix: add translation string card appearance

niekvanstaveren:
I guess the lint has updated as well, but main change on line 205/207

### DIFF
--- a/src/extension/ext-raw.js
+++ b/src/extension/ext-raw.js
@@ -11,7 +11,7 @@ const navigationOptions = [
   { value: 'expandAll', translation: 'Object.OrgChart.ExpandAll' },
 ];
 
-const bordersActive = data =>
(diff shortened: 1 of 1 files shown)


src/extension/ext-raw.js
```diff
@@ -11,7 +11,7 @@ const navigationOptions = [
   { value: 'expandAll', translation: 'Object.OrgChart.ExpandAll' },
 ];
 
-const bordersActive = data =>
+const bordersActive = (data) =>
```
haxxmaxx, line 14: my linter does this as well, do we want it with brackets??

niekvanstaveren: I just go with what the linter does. It is updated every week I think, so probably this is coming from one of those lint updates

PR #154, opened 2020-06-11
fix: render snapshot without viewstate

niekvanstaveren:


haxxmaxx:
Does this have a jira issue?

niekvanstaveren:
> Does this have a jira issue?

It does, but we do not reference the Jira issues in qlik-oss

PR #155, opened 2020-06-11
feat: add version to qext file

niekvanstaveren:
This adds/bumps the version number in the qext file. Version number is necessary when upgrading sense client, otherwise the extension won't be present after the upgrade.

We should check if a version is necessary both on the old and the new extension and if we need some patches to be created, but this is a first step

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,8 @@
     "start": "nebula serve --type sn-org-chart",
     "test:unit": "aw -c aw.config.js",
     "test:integration": "aw puppet --glob 'test/integration/**/*.spec.js' --chrome.headless true --chrome.slowMo 10",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -35,7 +35,8 @@
     "start": "nebula serve --type sn-org-chart",
     "test:unit": "aw -c aw.config.js",
     "test:integration": "aw puppet --glob 'test/integration/**/*.spec.js' --chrome.headless true --chrome.slowMo 10",
-    "prepublishOnly": "rm -rf dist && npm run build"
+    "prepublishOnly": "rm -rf dist && npm run build",
+    "version": "node ./scripts/version-qext.js && git add -A"
```
niekvanstaveren, line 39: Potentially we can add a postversion script as well that pushes the changes and tag

PR #198, opened 2021-04-06
fix: bugfix

eliseeborn:


Caele (approved):

eliseeborn:
did some testing, and there might still be an issue in the tooltip (or yarn link isn't working), looking into it a bit more

niekvanstaveren (approved):
Looks like everything in the card and tooltip is done now, great work!

PR #206, opened 2021-04-20
fix: patch #198 into April 2020

eliseeborn:


src/card/card.js
```diff
@@ -23,15 +24,15 @@ export default (data, cardStyling, sel, selectionState) => {
   const topColor = colorUtils.getDarkColor(backgroundColor);
   const fontColor = getFontColor(cardStyling, backgroundColor);
```
niekvanstaveren, line 25: This is not encoded now

src/card/card.js
```diff
@@ -1,10 +1,11 @@
...
+import encodeUtils from '../utils/encoder';
 
 export function getBackgroundColor(data, cardStyling) {
   let color = cardStyling.backgroundColor;
   if (data.attributes && data.attributes.color) {
     const resolvedColor = colorUtils.resolveExpression(data.attributes.color);
-    color = resolvedColor !== 'none' ? resolvedColor : color;
+    color = resolvedColor !== 'none' ? resolvedColor : encodeUtils.encodeCssColor(color);
   }
   return color;
```
niekvanstaveren, line 10: This is not encoded

src/card/card.js
```diff
@@ -1,10 +1,11 @@
 import colorUtils from '../utils/color-utils';
+import encodeUtils from '../utils/encoder';
 
 export function getBackgroundColor(data, cardStyling) {
   let color = cardStyling.backgroundColor;
   if (data.attributes && data.attributes.color) {
     const resolvedColor = colorUtils.resolveExpression(data.attributes.color);
-    color = resolvedColor !== 'none' ? resolvedColor : color;
+    color = resolvedColor !== 'none' ? resolvedColor : encodeUtils.encodeCssColor(color);
```
niekvanstaveren, line 8: encodeCssColor does not include all options. Instead of changing the encoder returning the result on line 7 would also be okay

src/card/card.js
eliseeborn: the font color is only set in the code, without user input

src/card/card.js
eliseeborn: I thought resolveExpression would be good enough in this case, but I did miss the color set before the if-statement

src/card/card.js
niekvanstaveren: That is only for default font color. What about this:

image

eliseeborn: you're right, I didn't read the code correctly there

src/card/card.js
eliseeborn: should I just copy the resolveColor function from master?

niekvanstaveren: That should work!

niekvanstaveren (approved):
Yes looks good, great job!

PR #240, opened 2021-11-24
fix: Remove case for IE11 transitions

haxxmaxx:


niekvanstaveren:
@haxxmaxx  We should probably merge this at some moment right?

haxxmaxx:
> @haxxmaxx We should probably merge this at some moment right?

yes, I just haven't tested it properly. It should be risk free, but need to test it quickly since it is so old at this point

PR #241, opened 2021-11-25
docs: add api spec

eliseeborn:
I think this is my first time using scriptappy but it went ok


src/object-properties.js
```diff
@@ -27,77 +31,117 @@ const properties = {
...
        * @type {object}
        */
       color: {
+        /**
+         * Index of theme color
+         * @type {number}
+         */
         index: -1,
+        /**
+         *  Color defined by rgb, argb or html
```
haxxmaxx, line 64: Maybe others work as well, but we generally want this to be a hex color

src/object-properties.js
```diff
@@ -27,77 +31,117 @@ const properties = {
...
        */
       colorExpression: '',
     },
     /**
+     * Holds border details
      * @type {object}
      */
     border: {
       /**
+       * Shop top border
```
haxxmaxx, line 113: show?

haxxmaxx:
Which of these properties are optional? I ran into the problem for the action button that if you want to change like one property for the styling, you override all the others and there are no sanity checks for that in the code, so you have to define the defaults again... What is the case for org chart? Maybe no need to change it, but it is a nicer experience for mashup developers

haxxmaxx (approved):
LGTM! I'll let @cbt have a look as well

src/object-properties.js
```diff
@@ -27,13 +31,60 @@ const properties = {
...
+ * Holds styling options
+ * @typedef {object} Style
+ * @property {FontColor=} fontColor - Color of the text, by default #484848
+ * @property {Background=} backgroundColor - Color of the background, by default #e6e6e6
+ * @property {Border=} border - Styling for border, by default #737373
+ */
+
+/**
+ * @typedef {object} FontColor
+ * @property {('auto'|'colorPicker'|'byExpression')=} colorType - How the font color is defined, defaults to auto
```
cbt1, line 55: For properties with primitive values and where the value is optional. You can specify a default with something like this I believe:
`@property {('auto'|'colorPicker'|'byExpression')} [colorType='auto']`

That would should generate a default value in the spec as well so no need to type the default value in the property description. For the case with objects as props I'll have to get back to you on how to best define that.

api-specifications/properties.json
```diff
@@ -0,0 +1,234 @@
...
+      "kind": "object",
+      "entries": {
+        "top": {
+          "description": "top - Set to true to show thicker top border",
+          "optional": true,
+          "defaultValue": true,
+          "type": "boolean"
+        },
+        "fullBorder": {
+          "description": "fullBorder - Set to true to show full border, default is false",
```
cbt1, line 59: The description looks odd with the property name there as well. Try changing the jsdoc definition to:
`@property {boolean} [fullBorder=false] Set to true to show full border`

cbt1:
FYI you can run the dev-portal locally the test how the API spec appears in the dev-portal. To get a better sense if it looks alright or not.

src/object-properties.js
haxxmaxx: That is really useful. Would it still be considered a default value even if there is no real default value on the properties per se, but handled with a fallback value in the styling logic? Asking since I wonder if this would work for the table as well

cbt1: I would say the default value is there to indict to the user how the behaviour will be, given that they do not provide a value. So it should not matter how it's implemented. As long as the default value results in the same behaviour as when they do set an explicit value.

eliseeborn:
Union types don't seem to work on base-level properties :/

Screenshot 2021-12-14 at 12 48 34
Screenshot 2021-12-14 at 12 48 11

Otherwise I think it's looking pretty good once I remove the property names from the description

eliseeborn:
or no, because properties should be one whole object, which would probably solve the above problem - might just be that I have some issue in my scriptappy setup

api-specifications/properties.json
```diff
@@ -0,0 +1,246 @@
...
+              "value": "'regular'"
+            },
+            {
+              "kind": "literal",
+              "value": "'scroll'"
+            },
+            {
+              "kind": "literal",
+              "value": "'free'"
+            }
```
niekvanstaveren, line 41: These are not the correct options right? I think there is free and expandAll mode. We have talked about different modes, but never implemented any

package.json
```diff
@@ -33,6 +33,7 @@
     "sense": "nebula sense --ext ./src/extension/ext-raw.js --meta ./src/extension/meta.json",
     "lint": "eslint src",
     "start": "nebula serve --type sn-org-chart",
+    "spec": "scriptappy-from-jsdoc -c ./spec-configs/props.conf.js",
```
niekvanstaveren, line 36: Does this need to run automatically? E.g. when version is bumped?

niekvanstaveren (commented):
Good stuff!

package.json
eliseeborn: I'll do that in a separate PR

cbt1 (approved):
LGTM

src/object-properties.js
```diff
@@ -56,4 +68,49 @@ const properties = {
...
+ * @property {PaletteColor=} color - Color defined by index or hex code, needed if colorType is colorPicker
+ * @property {ColorExpression=} colorExpression - Color defined by expression, needed if colorType is byExpression
+ */
+
+/**
+ * @typedef {object} Background
+ * @property {('auto'|'colorPicker'|'byExpression')=} [colorType='colorPicker'] How the font color is defined
+ * @property {PaletteColor=} color - Color defined by index or hex code, needed if colorType is colorPicker
+ * @property {ColorExpression=} colorExpression - Color defined by expression, needed if colorType is byExpression
+ */
```
haxxmaxx, line 91: These two are identical, could be one generic color type

eliseeborn: They are basically the same, but they have different default values for colorType

haxxmaxx: you're right, my bad! It still says `font` on line 88 though

PR #242, opened 2021-11-26
feat: make rendering testing work with enigma-mocker and test fixtures

LiKang6688:
Nebula serve using mocked Qlik core engine (Enigma mocker) to render the visualization charts based on the testing fixture file

Testing cases using Puppeteer capture screenshots of those rendered charts and matches against the baseline to ensure our charts look as intended and present clients with expected visuals.

Visual regression Testing flow:

1. Start Nebula serve and launch Puppeteer
2. Iterate testing fixture files
3. Create test case per testing fixture file
4. Render chart based on testing fixture file in Nebula serve using Enigma mocker
5. Puppeteer capture screenshot
6. Compare screenshot with baseline image


package.json
```diff
@@ -35,6 +35,7 @@
     "start": "nebula serve --type sn-org-chart",
     "test:unit": "aw -c aw.config.js",
     "test:integration": "aw puppet --glob 'test/integration/**/*.spec.js' --chrome.headless true --chrome.slowMo 10",
+    "test:rendering": "aw puppet -c aw.ren.config.js --testExt '*.spec.js' --glob 'test/rendering/**/*.spec.js'",
```
niekvanstaveren, line 38: Why are we still using after work for this? Can't we fix both of them at the same time?

LiKang6688: Will be a new test-assistant in @nebula.js/test-utils to replace after work which sets up the puppeteer and does other repetitive work. but that is not ready yet.

haxxmaxx:
Same renaming should be done here as my comments on https://github.com/qlik-oss/sn-action-button/pull/253

haxxmaxx (approved):

PR #244, opened 2021-11-29
refactor: move defaults to ext-raw.js

eliseeborn:


src/style-defaults.js
```diff
@@ -0,0 +1,11 @@
+export default {
+  FONT_COLOR_TYPE: 'auto',
+  FONT_COLOR_DARK: { index: -1, color: '#484848' },
+  FONT_COLOR_LIGHT: { index: -1, color: '#e6e6e6' },
+  BACKGROUND_COLOR_TYPE: 'colorPicker',
+  BACKGROUND_COLOR: { index: -1, color: '#e6e6e6' },
```
haxxmaxx, line 6: Has this always been a hard-coded color and not a palett color?

haxxmaxx: ahh of course, it is grey 👍

haxxmaxx (approved):
LGTM, as long as all defaults are covered :) I did the same with like empty strings, no need to have those as a constant

PR #250, opened 2022-01-25
fix: empty style object

eliseeborn:


cbt1:
Could this also been an issue on the other chart where default styling was removed? If not, why is that?

eliseeborn:
> Could this also been an issue on the other chart where default styling was removed? If not, why is that?

I think you mean the button, and in that case the default styling wasn't fully removed in the end.

src/object-properties.js
```diff
@@ -23,7 +25,8 @@ const properties = {
    */
   footnote: '',
   /**
-   * @type {'regular'|'scroll'|'free'}
+   * How the org chart is navigated
```
cbt1, line 28: This is a spec change and is best to do in separate PR. It also looks like a breaking change?

eliseeborn: It's been this way in the code for the whole time, just documented wrong, but it should be in a different PR nevertheless 👍

cbt1: Just remember that it will be breaking change from API Gov perspective, so once you do a new release make sure to those spec changes approved.

cbt1 (approved):

PR #253, opened 2022-02-03
chore: update build and release processes

eliseeborn:
mostly just checked what @haxxmaxx had done in action button

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@nebula.js/sn-org-chart",
   "version": "0.16.0",
-  "description": "",
+  "description": "Org chart supernova",
   "license": "MIT",
   "author": "QlikTech International AB",
   "keywords": [
@@ -20,8 +20,10 @@
     "url": "https://github.com/qlik-oss/sn-org-chart.git"
   },
   "files": [
-    "assets",
-    "dist"
+    "dist",
+    "api-specifications",
+    "core",
+    "sn-org-chart-ext"
   ],
   "engines": {
     "node": ">=16"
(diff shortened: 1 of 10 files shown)


haxxmaxx:
You need to add the missing environment variables to circle. You can import them from another project if you have access to that (I think you need to be admin). Have a look otherwise I can help you

eliseeborn:
> You need to add the missing environment variables to circle. You can import them from another project if you have access to that (I think you need to be admin). Have a look otherwise I can help you

I don't seem to have the right access, so some help would be nice :)

haxxmaxx:
The only one missing was the NPM_TOKEN so I added that. I thought that was why it was failing, but it is something else

package.json
```diff
@@ -71,13 +77,15 @@
     "eslint-config-airbnb-base": "14.2.1",
     "eslint-plugin-import": "2.25.4",
     "eslint-plugin-mocha": "9.0.0",
+    "fs-extra": "^10.0.0",
```
cbt1, line 80: Should be looked to version `10.0.0`

package.json
```diff
@@ -71,13 +77,15 @@
...
     "husky": "4.3.8",
     "jest": "^27.4.7",
     "jest-junit": "^13.0.0",
     "less": "^4.1.2",
     "lint-staged": "11.2.6",
     "mocha-junit-reporter": "2.0.2",
     "shx": "0.3.4",
+    "yargs": "^17.2.1",
```
cbt1, line 88: Should be looked to version `17.1.1`. Don't forget to run `yarn` after so that the `yarn.lock` file gets updated.

haxxmaxx:
The files in the `tools` dir are copy pasted right?

eliseeborn:
> The files in the `tools` dir are copy pasted right?

yes

haxxmaxx (approved):
LGTM

cbt1 (approved):

PR #258, opened 2022-02-08
chore: make scripts executable

eliseeborn:
change file permissions so they can run on circleci


cbt1 (commented):
Its safest to follow "principle of least privilege" here. So only the prepack script needs execution rights if I understand this correctly?

cbt1 (approved):

PR #283, opened 2023-03-06
chore: set unit test threshold

niekvanstaveren:


cbt1 (approved):
Set a coverage that matches what we have it your'e good to go

PR #285, opened 2023-03-30
chore: update localization files

eliseeborn:
based on how localization is handled in sn-table


jest.config.js
```diff
@@ -7,5 +7,5 @@ module.exports = {
     },
   },
   coveragePathIgnorePatterns: ['src/__tests__/default-orgchart-props.js', 'src/styles', 'src/extension', 'src/data.js'],
-  collectCoverageFrom: ['src/**'],
+  collectCoverageFrom: ['src/**', '!**/locales/**', '!**/all.json', '!**/*.mjs'],
```
cbt1, line 10: You could add these to `coveragePathIgnorePatterns` instead. Not sure why the build faild with lack of coverage. But possibly related to this.

.circleci/config.yml
```diff
@@ -94,6 +95,15 @@ jobs:
   #      - store_artifacts:
   #          path: ./sn-org-chart.zip
 
+  locale:
+    <<: *defaults
+    steps:
+      - attach_workspace:
+          at: ~/table
```
cbt1, line 102: Wrong workspace

yarn.lock
```diff
@@ -4,7 +4,7 @@
 
 "@after-work.js/aw@6.0.14":
   version "6.0.14"
-  resolved "https://registry.yarnpkg.com/@after-work.js/aw/-/aw-6.0.14.tgz#5bc86ab8189d08ab7752542a2ae3b137827b96c0"
```
cbt1, line 7: What is going on here? All deps have change registry?

jest.config.js
eliseeborn: I don't think it'll fix the coverage either way, this mostly is so that the coverage check doesn't throw a bunch of errors because it can't parse json-files

yarn.lock
eliseeborn: I added then removed some dependencies, seems like that really messed up the lock-file. deleting it and rerunning yarn didn't really seem to help 😐

cbt1: The seem to use the same registry now at least. But there a still a ton of changes for just adding `globby` as a dependency.

I'm guessing that is because the lock file had not been updated properly. So I think we can ignore it.

eliseeborn:
the locale check doesn't run correctly anymore apparently, I'll look into that tomorrow

PR #288, opened 2023-04-13
test: move integration and render tests to playwright

niekvanstaveren:


LiKang6688:
the rendering tests look good, but not the integration tests.

test/integration/rendering.spec.js
```diff
@@ -1,32 +1,33 @@
...
+  const content = '.sn-org-chart';
+
+  test.describe('from snapshot', () => {
+    test('basic example', async ({ page }) => {
+      await page.goto(route.renderSnapshot('basic'));
+      await page.waitForSelector(content, { visible: true });
       const elem = await page.$(content);
       const img = await elem.screenshot();
-      return expect(img).to.matchImageOf('snapshot-basic', artifacts);
+      expect(img).toMatchSnapshot('basic-snapshot.png');
```
LiKang6688, line 30: what's the different from the rendering tests?

niekvanstaveren: Not much, the integration tests are the snapshot tests. Not really clear this, but that was already the setup.
Should probably have only one entry (rendering)

test/rendering/snapshots.spec.js
```diff
@@ -0,0 +1,15 @@
...
+test.describe('should render', () => {
+  const content = '.sn-org-chart';
+
+  test.describe('from snapshot', () => {
+    test('basic example', async ({ page }) => {
+      await page.goto('/render?snapshot=basic', { waitUntil: 'networkidle' });
+      await page.waitForSelector('.sn-org-chart-snapshot', { visible: true });
+      const elem = await page.$(content);
+      const img = await elem.screenshot();
+      expect(img).toMatchSnapshot('basic-snapshot.png');
```
LiKang6688, line 12: I learned that toMatchSnapshot() now should be used for comparing text or arbitrary binary data
to produce and visually compare screenshots using await expect(page).toHaveScreenshot().
https://playwright.dev/docs/test-snapshots
I did that in https://github.com/qlik-oss/sn-table/pull/988/files#diff-d8d306b0b30eef9dd807a03675987a618b0b89a804d01988bd7dbc8afb215c8fR72

niekvanstaveren: Oh that is good to know. I'll update!

nebula.config.js
```diff
@@ -2,6 +2,7 @@
 module.exports = {
   build: {},
   serve: {
+    open: false,
```
LiKang6688, line 5: why do you want to open the app?

niekvanstaveren: I don't want to open. Personally I think it is annoying when it opens a new window every time you start the nebula server. Often I still have a page open actually. For running the integration tests locally it also does not make sense and currently they share the config

playwright.config.js
```diff
@@ -0,0 +1,14 @@
+const config = {
```
LiKang6688, line 1: it is good to add more here, like I did in https://github.com/qlik-oss/sn-table/pull/988/files#diff-f679bf1e58e8dddfc6cff0fa37c8e755c8d2cfc9e6b5dc5520a5800beba59a92R3

LiKang6688 (approved):
except for the configs, the rest look good to me

niekvanstaveren:
> except for the configs, the rest look good to me

As discussed on slack we will update this when the need comes

PR #293, opened 2023-04-19
chore: update lint and prettier to qlik config

niekvanstaveren:


.prettierignore
```diff
@@ -1,4 +1,18 @@
...
+CODEOWNERS
+LICENSE
+coverage
+core
 dist/
-
-*.html
\ No newline at end of file
+node_modules
+yarn.lock
```
cbt1, line 12: I think you should include the API spec file too. Generated files should never be "prettified" IMO

cbt1 (approved):
A minor comment only, LGTM

.prettierignore
niekvanstaveren: Yes!

PR #314, opened 2023-09-18
chore: update scriptappy

niekvanstaveren:
In preparation of the move to sn-gh-workflows


package.json
```diff
@@ -91,6 +91,8 @@
     "@playwright/test": "1.38.0",
     "@qlik/eslint-config": "0.4.18",
     "@qlik/prettier-config": "0.3.0",
+    "@scriptappy/cli": "^0.10.0",
```
cbt1, line 94: Pin versions

niekvanstaveren: Cannot believe you have to keep pointing this out 😊 
Maybe there is a way to have them pinned when installed actually, I will have a look around. Thanks!

cbt1: Maybe when you install an explicit version they get pinned? But yeah that would be nice to have since it easy to forget

cbt1 (approved):

package.json
niekvanstaveren: And of course it exists https://classic.yarnpkg.com/lang/en/docs/cli/add/#toc-yarn-add-exact-e

niekvanstaveren: Even better https://github.com/yarnpkg/yarn/issues/3045

PR #315, opened 2023-09-18
chore: move to sn-gh-workflows

niekvanstaveren:


.circleci/config.yml
```diff
@@ -1,183 +1,15 @@
...
-    name: Save yarn cache
-    paths:
-      - ~/.cache/yarn
-    key: yarn-packages-{{ .Branch }}-{{ checksum "yarn.lock" }}
-  - &filter_master
-    branches:
-      only: master
-  - &attach
-    at: ~/project
+version: 2.1
```
cbt1, line 1: Why an empty workflow?

cbt1:
You should make the `Build` action required once merge

cbt1 (approved):

.circleci/config.yml
niekvanstaveren: Because the circleci flow is required. We can switch that to the Build action once this is merged :)

PR #350, opened 2024-03-13
Styling panel for org-chart

ayakushi:
ORG-CHART STYLING PANEL

**GENERAL TAB**
1. Font family/font size/font styling/font color for title/subtitle/footer
2. Chart Background Color (single color or expression)

**CHART TAB**

Properties migrated from Property Panel to Styling Panel
1. Font color (Auto, Single color and By expression)
2. Card background color (Auto, Single color and By expression)
3. Card appearance (top bar, border and border color)

New Properties
1. Font family/font size for card title
2. Font family/font size for card body

image


nebula.config.js
```diff
@@ -3,6 +3,9 @@ module.exports = {
   build: {},
   serve: {
     open: false,
+    flags: {
+      SENSECLIENT_IM_5036_VIZBUNDLE_STYLING: true,
```
haxxmaxx, line 7: Is this flag for the org-chart styling only? If so, it would be better having ORG_CHART in the flag name. Maybe it is too late to change?

ayakushi: Yes, the flag is only for org-chart. Initially it was meant to include the variance waterfall styling also, but the scope was changed afterwards and I didn't update the name. I will change it to IM_5036_ORGCHART_STYLING and update qlik-sense-desktop and qlik-sense-server.

package.json
```diff
@@ -35,7 +35,9 @@
     "sense": "nebula sense --ext ./src/extension/ext-raw.js --meta ./src/extension/meta.json",
     "spec": "sy from-jsdoc -c ./spec-configs/props.conf.js",
     "start": "nebula serve --type sn-org-chart",
+    "start:mfe": "nebula serve --mfe --type sn-org-chart",
     "test:e2e": "playwright test",
+    "test:rendering": "playwright test --config=./playwright.config.js",
```
niekvanstaveren, line 40: Why add this? This should do the same as `test:e2e` (which is used in the CI pipeline). The config parameter is the default

nebula.config.js
ayakushi: Hi, I've created a new PR https://github.com/qlik-oss/sn-org-chart/pull/354 where the feature flag has been changed and also following your suggestion about splitting I've only included the property panel and property migration.

package.json
ayakushi: You are right, correction made in this new PR https://github.com/qlik-oss/sn-org-chart/pull/354 where I've split the code to include only property panel and property migration.

ayakushi:
This PR will be split in 2 parts to facilitate code review. New PR https://github.com/qlik-oss/sn-org-chart/pull/354

PR #354, opened 2024-03-19
Styling panel for org-chart

ayakushi:
ORG-CHART STYLING PANEL

**GENERAL TAB**
1. Font family/font size/font styling/font color for title/subtitle/footer
2. Chart Background Color (single color or expression)

**CHART TAB**

Properties migrated from Property Panel to Styling Panel
1. Font color (Auto, Single color and By expression)
2. Card background color (Auto, Single color and By expression)
3. Card appearance (top bar, border and border color)

New Properties
1. Font family/font size for card title
2. Font family/font size for card body

image


src/extension/migration.js
```diff
@@ -0,0 +1,41 @@
...
+    border: {
+      top: undefined,
+      fullBorder: undefined,
+      colorType: undefined,
+      color: undefined,
+      colorExpression: undefined,
+    },
+  };
+
+  if (!properties.components || properties.components.length === 0) {
```
niekvanstaveren, line 30: What would happen when you reset the styling panel? That removes all the entries in the components array I think. Do you migrate again?

niekvanstaveren (commented):
So this PR adds the styling panel itself, but does not handle the new property location I guess? E.g. changing the values in the property panel does not actually do anything?

src/extension/migration.js
ayakushi: No, it will not migrate again. The reset will clean up the old styling and use the default values for the properties instead.

ayakushi: This PR only adds the styling panel and applies the migrated values. All the logic to apply the selections made on the styling panel will be included ini a second PR.

src/extension/ext-raw.js
```diff
@@ -15,262 +17,287 @@ const navigationOptions = [
...
+  return settings;
 };
+
+export default function extDef({ translator, flags, anything }) {
+  return {
+    definition: {
+      type: "items",
+      component: "accordion",
+      items: {
+        data: getData(flags),
```
niekvanstaveren, line 285: The getData function does not seem to have any inputs. Does not have to be a function I think

src/extension/migration.js
niekvanstaveren: Okay, that might give issues with backwards compatibility I guess. But sometimes that is not something we can go without.

src/object-properties.js
```diff
@@ -66,6 +66,7 @@ const properties = {
       colorExpression: "",
     },
   },
+  components: [],
```
niekvanstaveren, line 69: This is probably a good addition. However when we add the logic for resolving everything we need to make sure that we also add tests for a missing components array

niekvanstaveren (approved):
Looks good to me. Maybe good if @haxxmaxx has another look at it as well

src/extension/ext-raw.js
ayakushi: Right, this is the remanent after I split the styling and the images feature into different PRs. The flag will be needed for the image feature. I will remove it now an include it in upcoming PR.

src/object-properties.js
ayakushi: I will add that scenario to the migration test. But there is another unit test included in the second PR that take the missing components into consideration.

src/extension/migration.js
ayakushi: Yes, that's true. The styling defined with the styling panel will not be applied if we turn off the feature flag afterwards. PM approved with that caveat.

src/extension/styling-definitions/styling-definition.js
```diff
@@ -0,0 +1,220 @@
...
+    flags,
+    config: {
+      id: "orgChart",
+      paths: ["axis.label.name", "label.value"],
+    },
+  });
+
+  const bordersActive = (data) =>
+    (propertyResolver.getValue(data, "border.top") ?? DEFAULTS.BORDER_TOP) ||
+    (propertyResolver.getValue(data, "border.fullBorder") ?? DEFAULTS.BORDER_FULL);
```
haxxmaxx, line 25: If `propertyResolver.getValue(data, "border.top")` returns something falsey, doesn't the function always return `DEFAULTS.BORDER_TOP`? So the second part after `||` isn't needed?

src/extension/styling-definitions/styling-definition.js
```diff
@@ -0,0 +1,220 @@
...
+                translation: "Object.OrgChart.TopBar",
+                defaultValue: DEFAULTS.BORDER_TOP,
+              },
+              fullBorder: {
+                component: "checkbox",
+                ref: "border.fullBorder",
+                translation: "properties.border",
+                defaultValue: DEFAULTS.BORDER_FULL,
+              },
+              fontColorWrapperItem: {
```
haxxmaxx, line 174: Doesn't really matter, but I guess it should be called something with border

src/extension/migration.js
```diff
@@ -0,0 +1,41 @@
...
+    Object.assign(dataTemplateLabel.label.value, properties.style?.fontColor);
+    Object.assign(dataTemplateBackground.backgroundColor, properties.style?.backgroundColor);
+    Object.assign(dataTemplateBorder.border, properties.style?.border);
+    const labelComponent = { key: "label", ...dataTemplateLabel };
+    const bgColorComponent = { key: "backgroundColor", ...dataTemplateBackground };
+    const borderComponent = { key: "border", ...dataTemplateBorder };
+    if (!properties.components) properties.components = [];
```
haxxmaxx, line 37: Looks like you can early return in the beginning of the function if `components` doesn't exist?

haxxmaxx (commented):
Couple minor comments, otherwise looks good!

src/extension/styling-definitions/styling-definition.js
ayakushi: The ?? is returning the first not null/undefined value, so it will still return false if the value for border.top is false https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

src/extension/styling-definitions/styling-definition.js
ayakushi: Yes, I probably forgot to rename after copy paste. I will fix it.

src/extension/migration.js
ayakushi: You are right, I will move the condition to the beginning.

src/extension/styling-definitions/styling-definition.js
haxxmaxx: ahh, I didn't know that. I thought it was for false as well, good to know

cbt1: Do you need the propertyResolver here? Would `data?.border?.top` not be enough?

cbt1: Also to clarify. The intention is that if the top border is `0`, it should fallback to using the full border?

cbt1 (commented):
I'm fairly sure the `propertyResolver` is legacy code that exist because the nullish coalescing (??) operator and the optional chaining (?.) operator did not exist a few years ago.

But since we can use that now should aim not use it any more and instead use `??` and `?.` when accessing properties.

src/extension/migration.js
```diff
@@ -0,0 +1,41 @@
...
+      color: undefined,
+      colorExpression: undefined,
+    },
+  };
+
+  if (!properties.components || properties.components.length === 0) {
+    Object.assign(dataTemplateLabel.label.value, properties.style?.fontColor);
+    Object.assign(dataTemplateBackground.backgroundColor, properties.style?.backgroundColor);
+    Object.assign(dataTemplateBorder.border, properties.style?.border);
+    const labelComponent = { key: "label", ...dataTemplateLabel };
```
cbt1, line 34: Do you really need to create another object here just to assign the key? Why not juse add the key when `dataTemplateLabel` is created?

src/extension/migration.js
```diff
@@ -0,0 +1,41 @@
+export default function migrateStyle(properties) {
+  const dataTemplateLabel = {
+    label: {
+      value: {
+        colorType: undefined,
```
cbt1, line 5: Why are these properties here if they are `undefined`?

ayakushi: I've made some changes to migration to fix all the problems mentioned above.

src/extension/migration.js
```diff
@@ -0,0 +1,32 @@
...
+  };
+
+  const dataTemplateBorder = {
+    key: "border",
+    border: {
+      ...properties.style?.border,
+    },
+  };
+
+  if (!properties.components) properties.components = [];
```
cbt1, line 27: It looks much better now!

But this can still be improved a little by removing the if statement and assigning components array directly: `properties.components = [dataTemplateLabel, dataTemplateBackground, dataTemplateBorder];`

src/extension/migration.js
```diff
@@ -0,0 +1,32 @@
+export default function migrateStyle(properties) {
+  if (properties.components && properties.components.length > 0) return properties;
```
cbt1, line 2: This can actually be shortend to `if (properties.components?.length > 0)...`

cbt1 (approved):
LGTM. Just had some final minor comments

src/extension/styling-definitions/styling-definition.js
ayakushi: The bordersActive function is only used to show/hide the color dropdown selector for the border. The selector will be displayed if at least one of them is true (top or fullBorder). 
If top is 0, it doesn't affect the full border value. They are independent of each other.

cbt1: I see. I was not expecting them to be booleans.

